PR #5, opened 2016-12-09
Use UTC timezone on _parse_time

FooBarQuaxx:
Fixes test error on non-UTC timezone system like mine (AST which is UTC+3:00). Here the output of the ./runtests.sh before this patch:
```
.....
======================================================================
FAIL: test_decode_linux (tests.test_ftp_parse.TestFTPParse)
----------------------------------------------------------------------
  File "/Users/xxx/Code/pyfilesystem2/tests/test_ftp_parse.py", line 108, in test_decode_linux
    self.assertEqual(parsed, expected)
AssertionError: Lists differ: [{'basic': {'is_dir': True, 'name': 'debian'[1530 chars]t'}}] != [{'access': {'permissions': ['g_r', 'g_w', '[1530 chars]t'}}]

First differing element 0:
{'basic': {'is_dir': True, 'name': 'debian'[282 chars]an'}}
{'access': {'permissions': ['g_r', 'g_w', '[282 chars]an'}}

Diff is 3098 characters long. Set self.maxDiff to None to see it.

======================================================================
FAIL: test_parse_time (tests.test_ftp_parse.TestFTPParse)
----------------------------------------------------------------------
  File "/Users/xxx/Code/pyfilesystem2/tests/test_ftp_parse.py", line 15, in test_parse_time
    142128000.0
AssertionError: 142117200.0 != 142128000.0
.....
Ran 651 tests in 1.248s

FAILED (failures=2)
```
Note that 142128000.0 - 142117200.0 = 3hours.

### DIFF
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 
 import re
 import time
+import calendar
 
 from .enums import ResourceType
 from .permissions import Permissions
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks. I don't think I can use that fix though. It would fix the tests, but there is still a flaw there.

The problem is that method parses a time from an ftp server, which doesn't include timezone information. In order to fix that properly I would need to get the timezone from the server. I'm going to file an issue, to remind me figure out how to do that.

PR #16, opened 2017-01-16
Add support for tar (TarFS) and tests

althonos:
This PR contains edits to support Tar archives (through python **tarfile** module).

I tried to be consistent with the ZipFS implementation, although since tar archives are much more Linux-friendly, there was no need for the mirrored MemoryFS for ReadTarFS.

For the tar formats, supported compressions are 'gz', 'bz2', and 'xz' for Python3 only. When a TarFS is created, the compression is guessed from the file extension, but if needed it can be given with the `compression` argument as well.

I also added tests, based on the zipfs tests, and tried to stay as clear as possible. Tell me if you have any questions.

### DIFF
--- a/fs/compress.py
+++ b/fs/compress.py
@@ -11,10 +11,13 @@ from __future__ import unicode_literals
 from datetime import datetime
 import time
 import zipfile
+import tarfile
 
 import six
 
+from .enums import ResourceType
 from .path import relpath
+from .time import datetime_to_epoch
 from .errors import NoSysPath
 from .walk import Walker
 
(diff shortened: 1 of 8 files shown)


althonos:
I've got to fix Py3.3 though, so take your time :smile:

althonos:
Pfew, here we are !
I added two more asserts to archive_test_case:
* assert that calling listdir on a file raises DirectoryExpected instead of ResourceNotFound
* assert that calling openbin on a directory raises FileExpected instead of ResourceNotFound

I fixes ZipFS since it always raised ResourceNotFound, so now the coverage is back up again :)

willmcgugan:
This is awesome, thanks!

PR #20, opened 2017-02-04
tweaks to tarfs, mostly formatting and style

willmcgugan:


fs/compress.py
```diff
@@ -152,26 +160,21 @@ def write_tar(src_fs,
             tar_info = tarfile.TarInfo(tar_name)
 
             if info.has_namespace('stat'):
-                mtime = info.get('stat', 'st_mtime', None)\
-                                    or time.time()
+                mtime = info.get('stat', 'st_mtime', current_time)
```
willmcgugan: This is so that if the time has to be assumed, it is consistent. Other wise times will be different by some milliseconds. Quite what the impact would be, I'm not sure.

fs/compress.py
```diff
@@ -152,26 +160,21 @@ def write_tar(src_fs,
...
-                                     'uname':'user',
-                                     'gname':'group'}.items():
+            for tarattr, infoattr in tar_attr:
                 if getattr(info, infoattr) is not None:
-                    setattr(tar_info, tarattr,
-                            getattr(info, infoattr))
+                    setattr(tar_info, tarattr, getattr(info, infoattr))
 
-            tar_info.mode = getattr(info.permissions, 'mode', 420)
+            tar_info.mode = getattr(info.permissions, 'mode', 0o420)
```
willmcgugan: This should be octal, right?

fs/tarfs.py
```diff
@@ -64,9 +64,9 @@ class TarFS(WrapFS):
 
     _compression_formats = {
         #FMT    #UNIX      #MSDOS
-        'xz':  ('.tar.xz',  '.txz'),
+        'xz': ('.tar.xz', '.txz'),
```
willmcgugan: I have a linter in Sublime Text which complains about extra spaces.

fs/tarfs.py
```diff
@@ -225,10 +225,9 @@ def getinfo(self, path, namespaces=None):
                 raise errors.ResourceNotFound(path)
 
 
-            raw_tar_info = member.get_info(
-                *([self.encoding, None]
-                if six.PY2 else [])
-            )
+            raw_tar_info = member.get_info(*(
+                [self.encoding, None] if six.PY2 else []
```
willmcgugan: This was so that coverage isn't different between Py2 and 3 right? I do the same thing.

fs/tarfs.py
```diff
@@ -278,14 +277,14 @@ def listdir(self, path):
...
-            if dirname(member.path)==path
+            if dirname(member.path) == path
         ]
 
     def makedir(self, path, permissions=None, recreate=False):
         self.check()
         raise errors.ResourceReadOnly(path)
 
-    def openbin(self, path, mode="r", buffering=-1, **kwargs):
+    def openbin(self, path, mode="r", buffering=-1, **options):
```
willmcgugan: Just to match the signature in base.

fs/tarfs.py
althonos: Yes, fonction signature is not the same between Py2 and Py3, so this makes it consistent.

PR #27, opened 2017-03-10
WebDAV File System

SerSamgy:
Add WebDAV file system support

### DIFF
--- a/fs/webdavfs.py
+++ b/fs/webdavfs.py
@@ -30,16 +30,28 @@ class WebDAVFile(object):
         self.path = path
         self.res = self.fs.get_resource(self.path)
         self.mode = mode
+        self.data = self._get_file_data()
 
         self._lock = threading.RLock()
         self.pos = 0
         self.closed = False
 
         if 'a' in mode:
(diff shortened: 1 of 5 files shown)


willmcgugan:
Thanks for the contribution. This looks awesome, but I have a concern. I'm trying to keep the dependancies to a minimum, and it looks like webdavclient requires lxml, which is a pain to install on a lot of systems. lxml won't install on Raspberry Pi without some effort, for example.

I think it may be best to keep it as a separate module. I think there are two options; A) manage releases yourself and keep it under your repos or B) move it to the pyfilesystem organization and release it as a namespace package, so you can do `pip install fs.webdav`.

I have no particular preference. If A) I would link to it prominently in the docs and the readme.

What do you think?

willmcgugan:
:+1:

zopyx:
This PR can be closed. It will be replaced soon with a new implementation that will be based on the requests module instead of the pycurl dependency.

willmcgugan:
Ok. Let me know if you need any help with that.

PR #29, opened 2017-03-12
The document is translated into Chinese

fbigun:
Chinese localization。


fbigun:
Chinese localization

http://pyfilesystem2-zh.readthedocs.io/zh_CN/latest/

willmcgugan:
This is great. But I can't merge as is, since it overwrites the English originals. If you could add the translations along side the original text, I'd be happy to merge that.

fbigun:
I do not know where the translated content is specifically placed.

PR #30, opened 2017-03-13
Fixed UnicodeEncodeError when creating OSFS from non-ascii path

vilius-g:
It seems that os.pathconf has issues with unicode paths in Python 2 as it tries to decode to ascii despite locale settings. It string is encoded to bytes first, then everything works fine.

Tested on linux-3.13.0-110-generic/Python 2.7.6.


willmcgugan:
Thanks.

PR #43, opened 2017-05-07
Use PEP-496 syntax to resolve #41

### LINKED ISSUE #41: Enum34 is not compatible with Python 3.6
We've been hitting a bug with PyFilesystem2 in conjunction with Python 3.6. Depending on how the Python path is set up we are unable to use the enum features introduced in 3.6 if we also install PyFilesystem2.

The problem is that [enum34 does not support Python 3.6](https://bitbucket.org/stoneleaf/enum34/issues/19/enum34-isnt-compatible-with-python-36) and should therefore not be installed if the version used is greater than Python 3.4.

This is what you can use to test it locally:

```bash
virtualenv -p python3.6 venv
venv/bin/pip install fs
PYTHONPATH='./venv/lib/python3.6/site-packages' venv/bin/python -c "from enum import IntFlag"
```

Can this be changed? An option might be to use [enum-compat](https://pypi.python.org/pypi/enum-compat) which installs enum34 conditionally. I see there's also some sort of version-specific dependency in your setup.py though I haven't ever encountered that before.

althonos:
I just made the `enum32` a conditional requirement depending on the python version ;  this should be enough to fix #41.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 appdirs==1.4.0
 enum34==1.1.6 ; python_version < '3.4'
 pytz
-scandir==1.3
+scandir==1.3 ; python_version < '3.5'
 setuptools
 six==1.10.0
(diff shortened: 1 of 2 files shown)


willmcgugan:
Brilliant. Thanks.

PR #44, opened 2017-05-08
Make scandir an optional dependency

althonos:
Hi,
I just noticed while trying to install `fs` on the `alpine-python` Docker image that it had `scandir` as a hard requirement. Since `scandir` is a C extension, this meant that **1.** implementations other than CPython couldn't use *Pyfilesystem2*, and that **2.** it required C compilation, so minimal builds needed to have both a C compiler and Python headers installed, both of which can weigh a lot. 

With the edits in that PR, the `scandir` module will be used if it can be found (either `scandir` is installed or the version is superior to **3.5**). In other cases, plain `os.listdir` module will be used.

### DIFF
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -367,73 +367,68 @@ class OSFS(FS):
                         os.utime(sys_path, (accessed, modified))
 
 
-
-    def _scandir_c(self, path, namespaces=None):
-        self.check()
-        namespaces = namespaces or ()
-        _path = self.validatepath(path)
-        sys_path = self._to_sys_path(_path)
-        with convert_os_errors('scandir', path, directory=True):
-            for dir_entry in scandir(sys_path):
-                info = {
-                    "basic": {
-                        "name": dir_entry.name,
-                        "is_dir": dir_entry.is_dir()
+    if scandir:
+        def _scandir(self, path, namespaces=None):
+            self.check()
+            namespaces = namespaces or ()
+            _path = self.validatepath(path)
+            sys_path = self._to_sys_path(_path)
+            with convert_os_errors('scandir', path, directory=True):
+                for dir_entry in scandir(sys_path):
+                    info = {
+                        "basic": {
+                            "name": dir_entry.name,
+                            "is_dir": dir_entry.is_dir()
+                        }
                     }
-                }
-                if 'details' in namespaces:
-                    stat_result = dir_entry.stat()
-                    info['details'] = \
-                        self._make_details_from_stat(stat_result)
-                if 'stat' in namespaces:
-                    stat_result = dir_entry.stat()
-                    info['stat'] = {
-                        k: getattr(stat, k)
-                        for k in dir(stat) if k.startswith('st_')
-                    }
-                if 'access' in namespaces:
-                    stat_result = dir_entry.stat()
-                    info['access'] =\
-                        self._make_access_from_stat(stat_result)
+                    if 'details' in namespaces:
+                        stat_result = dir_entry.stat()
+                        info['details'] = \
+                            self._make_details_from_stat(stat_result)
+                    if 'stat' in namespaces:
+                        stat_result = dir_entry.stat()
+                        info['stat'] = {
+                            k: getattr(stat, k)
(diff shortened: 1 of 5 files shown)


fs/osfs.py
```diff
@@ -393,6 +398,43 @@ def _scandir(self, path, namespaces=None):
...
+            for entry_name in os.listdir(sys_path):
+                entry_path = os.path.join(sys_path, entry_name)
+                info = {
+                    "basic": {
+                        "name": entry_name,
+                        "is_dir": os.path.isdir(entry_path),
+                    }
+                }
+                if 'details' in namespaces:
+                    stat_result = os.stat(entry_path)
```
willmcgugan: Could you cache this `os.stat` call?  It could potentially be called several time. And perhaps you can use the `stat` structure rather than `isdir` on L412, because that will do a stat under the hood.

fs/osfs.py
```diff
@@ -363,7 +366,9 @@ def setinfo(self, path, info):
                     with convert_os_errors('setinfo', path):
                         os.utime(sys_path, (accessed, modified))
 
-    def _scandir(self, path, namespaces=None):
+
+
+    def _scandir_c(self, path, namespaces=None):
```
willmcgugan: I'd like to suggest a slightly different way of doing the conditional.

```python
try:
    import scandir
except ImportError:
    scandir = None

class OSFS:
    if scandir:
        def _scandir(self, path, namespaces=None, page=None):
            # with scandir implementation
    else:
        def _scandir(self, path, namespaces=None, page=None):
            # without scandir implementation
```

I think that's a little tidier than having separate `_scandir_c` and `scandir_py` functions. What do you think?

setup.py
```diff
@@ -37,7 +37,7 @@
     description="Python's filesystem abstraction layer",
     install_requires=REQUIREMENTS,
     extras_require={
-        ":python_version<'3.5'": ['scandir~=1.5'],
+        "scandir : python_version < '3.5'": ['scandir~=1.5'],
```
willmcgugan: I just merge a PR that might conflict with this. You might have to resolve it.

tests/test_osfs.py
```diff
@@ -94,3 +94,19 @@ def test_unicode_paths(self):
...
+
+
+class TestOSFS(OSFSTestBase, unittest.TestCase):
+    pass
+
+
+class TestOSFSScandir(OSFSTestBase, unittest.TestCase):
+
+    def test_scandir_purepython(self):
+        with mock.patch.object(self.fs, '_scandir', self.fs._scandir_py):
```
willmcgugan: I think I see why you have the 2 functions defined now. I think you can modif `tox.ini` file to have separate environments for with and without the scandir module. Otherwise its not going to include this in coverage.

willmcgugan:
Thanks! I concur. Only thing that concerns me is that scandir won't be installed on automatically in environments that it would easily install on. I guess that would just need to be documented.

fs/osfs.py
althonos: At first that's what I wanted to do, but for the sake of testing it's easier to have the two methods accessible at all times, since your way involves reloading the module while mocking an import to be able to access the other implementation. But I'll give it a shot anyway.

althonos:
I'll try to resolve the conflicts shortly.

For the automatic installation of scandir, I tried to search if there was a way to specify "default-but-not-mandatory" requirements, but no luck with that. At some point I thought restricting the installation of `scandir` to `sys_implementation == "cpython"`, but that would fail for CPython systems without Python headers.

The `setup.py` could also throw a warning if the version is `scandir` friendly, like:
```python
# setup.py

import sys
import warnings

if sys.implementation.name == "cpython" and sys.version < "3.5":
    try:
        import scandir
    except ImportError:
        warnings.warn("Consider installing scandir blah blah blah...")
```

althonos:
Done, you can merge when you want. I added another environment to `tox.ini`, and rolled back to defining `OSFS._scandir` directly depending on the availability of `scandir`.

willmcgugan:
Brilliant. Thanks.

Issue #50, opened 2017-06-13
Have objects returned by openbin inherit from IO.IOBase


althonos:
To support interfacing with the Python standard library, "file-like" objects should extend the [`IO.IOBase`](https://docs.python.org/3/library/io.html#io.IOBase) class (not really an abstract base class, but does nothing by itself), and the `fs.test.FSTestCases.openbin` updated accordingly with a `self.assertIsInstance(...)`.

Following classes are concerned, since the other filesystems already return a proper child object (`OSFS` and derived, `ZipFS`, `TarFS`)
* `fs.ftpfs.FTPFile`
* `fs.iotools.RawWrapper`
* `fs.memoryfs._MemoryFile`

I can provide a PR if you don't have the time to work on this @willmcgugan.

NB: neither `_MemoryFile` nor `FTPFile` implement `seekable`, `writeable` or `readable`, so they should be implemented to avoid returning `False` all the time like in the default `ÌOBase` implementation.

willmcgugan:
Sounds like a good idea. The differences between file-like objects has always been irksome. A PR would be great.

willmcgugan:
https://github.com/PyFilesystem/pyfilesystem2/pull/52

PR #52, opened 2017-06-14
Feat iobase (PR for #50)

althonos:
#### Lib
* made `FTPFile`, `_MemoryFile` and `RawWrapper` derive from `io.IOBase`
* removed `__enter__`, `_del__` and `__exit__` implementation as the base `io.IOBase` implementation returns `self` in `__enter__` and closes `self` in `__exit__` and `__del__` already.
* removed `RawWrapper` `next` and `__next__` as they only copied the base `io.IOBase` implementation
* implemented `writable` and `readable` depending on the mode of `FS.open` for `FTPFile` and `_MemoryFile`
* implemented `seekable` (always returning `True`) for all three classes

#### Tests
* Check file objects returned by `FS.open` and `FS.openbin` are instances of `io.IOBase`
* Check that `readable` and `writable` return the right value depending on the opening mode
* Check that file objects are not closed within a context block, then closed after
* Check that file objects are iterable over lines
* Fixed test as `f.fileno()` does not raise an `AttributeError` now (since it is implemented by `io.IOBase`) but an `io.UnsupportedOperation` if the subclass didn't reimplement that method.

### DIFF
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -132,8 +132,8 @@ class FTPFile(io.IOBase):
         return _repr.format(self.fs.ftp_url, self.path, self.mode)
 
     def close(self):
-        with self._lock:
-            if not self.closed:
+        if not self.closed:
+            with self._lock:
                 try:
                     if self._write_conn is not None:
                         self._write_conn.close()
@@ -252,7 +252,6 @@ class FTPFile(io.IOBase):
             if self._write_conn:
                 self._write_conn.close()
                 self._write_conn = None
-
 
 
 class FTPFS(FS):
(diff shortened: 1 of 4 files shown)


willmcgugan:
Thanks. Will review this soon.

althonos:
I have a question though, I am not sure I can see where `_MemoryFile` objects actually close their `_bytes_io` buffer. This does not really matter I suppose, since the buffer will be flushed nevertheless when the `_MemoryFile` is garbage collected, and the `BytesIO` garbage collected in turn.

willmcgugan:
> I have a question though, I am not sure I can see where _MemoryFile objects actually close their _bytes_io buffer. This does not really matter I suppose, since the buffer will be flushed nevertheless when the _MemoryFile is garbage collected, and the BytesIO garbage collected in turn.

IIRC the _MemoryFile references a BytesIO that is stored in the MemoryFS. So it isn't _MemoryFile's responsibility to close it.

But it doesn't look like MemoryFS ever explicitly closes anything. So you're right, it will only be freed when it is garbage collected. It would probably be better to close all the BytesIO objects when MemoryFS.close is called.

willmcgugan:
Thanks.

PR #64, opened 2017-08-12
Symlinks

willmcgugan:
Work in progress for reporting symlinks.


fs/info.py
```diff
@@ -71,6 +69,15 @@ def get(self, namespace, key, default=None):
         except KeyError:
             return default
 
+    def _require_namespace(self, namespace):
```
willmcgugan: @althonos This is a notable change. If a required namespace is not present, it raises a `MissingInfoNamespace` which is a subclass of `AttributeError`.

The motivation was the `islink` could have 3 return values; False, True and None. And I really didn't like that.

What do you think?

althonos: This looks like a good idea to me. Indeed the `isdir` and `isfile` worked because the `is_dir` attribute was in the *basic* namespace, but since the `is_link` cannot be forced into it, that's the only good solution for me. 

Although I'm not sure to understand: why 3 return values ?

willmcgugan: > Although I'm not sure to understand: why 3 return values ?

The `islink` method could return `None` to indicate the namespace was missing (in addition to False and True). The alternative would be to return `False` if the namespace didn't exist. I wasn't happy with either...

althonos: Well, you could have decided that:
* if the namespace is available, use it
* if it's not, try to use the `type` namespace and check the `type`
* if `type` is not available either, then the FS is not supporting symlinks, so return False.

But I get your point, this is not a satisfying solution.

PR #65, opened 2017-08-17
Update external FS page with personal contributions

althonos:
* Add `fs.smbfs` and `fs.sshfs`
* Add CI badge in filesystem section header


willmcgugan:
Cool, thanks. 

I'm actually planning to deprecate that section in the docs and provide a reference to an appropriate page in the [wiki](https://www.pyfilesystem.org) (which I'm slowly populating).

althonos:
Kudos about that, the website looks great !  (This is not something I can help with, alas ! 😄 )

PR #66, opened 2017-08-17
use FS URLs in multi and mount fs

willmcgugan:
- Adds openers for app urls, e.g. `userdata://myapp:willmcgugan:1.0`.
 - Allow MountFS and MultiFS to use FS Urls, e.g. `mount_fs.mount('foo', 'mem://')`


fs/opener/appfs.py
```diff
@@ -0,0 +1,55 @@
...
+        tokens = parse_result.resource.split(':', 3)
+        if len(tokens) == 2:
+            appname, author = tokens
+            version = None
+        elif len(tokens) == 3:
+            appname, author, version = tokens
+        else:
+            raise OpenerError(
+                'resource should be <appname>:<author> '
+                'or <appname>:<author>:<version'
```
althonos: Typo here

willmcgugan: :+1:

PR #68, opened 2017-08-31
Update tests, and fix discovered bugs

althonos:
Hi Will, I updated the tests in `fs.test` with some missing checks, and discovered some new bugs I will fix gradually in this PR.

### New tests
1. `FS.copy`, `FS.move`, `FS.copydir` and `FS.movedir` should raise `ResourceNotFound` when the source is nowhere to be found
2. `FS.copy` and `FS.move` should probably raise `FileExpected` when the source is a directory (not listed in the documentation, but makes sense I guess ?...)
3. `FS.movedir` and `FS.copydir` should raise `DirectoryExpected` when the source is a file
4. If a filesystem is advertising itself as *case insensitive*, check that's true
5. If a filesystem is advertising itself as supporting *unicode paths*, check that's true
6. Check that *file-like* objects returned by `FS.openbin`:
    * have a `seek` method that returns the new absolute position when called
    * have a `write` method that returns the number of bytes (in binary mode) / characters (in text mode) written
    * have a `truncate` method that returns the new length of the file

### Discovered bugs
* ~`FTPFile.seek`, `FTPFile.truncate` and `FTPFile.write` do not return anything~ *fixed*
* ~`FTPFS` crashes on unicode paths (`/földér`)~ *fixed, see below*
* ~`MemoryFile.write` and `MemoryFile.truncate` do not return anything~ *fixed*
* ~many filesystem's `move` and `copy` do not raise `FileExpected` when given a directory~ *fixed by fixing `FS.copy` and `FS.move`*

### About FTPFS
FTP servers supporting UTF-8 encoded paths display the feature `UTF8`. But the `ftplib.FTP` objects require to set the encoding *before* connecting to the server, or else the server and the client encoding will not be synchronised (client will send *utf-8* paths but server will treat them as *latin-1*). 

To fix this, the encoding is checked when accessing the `FTPFS.ftp` property for the first time. A mock connection is established to check for the `UTF8` feature ; then, a proper connection is established, with *utf-8* as the encoding if available, and *latin-1* otherwise. Then, any new ftp connection will use the previous encoding, so the overhead is limited.

### About test_ftpfs
I rewrote the class so that a new server is created once by class instead of before each test, which was a lot longer for nothing. I also made use of `pyftpdlib.test.FTPd` instead of externally calling the file and then *task-killing* the thread, which was a lot hackier. 

I still think there is an issue with the timeout of `FTPFS` objects, since the `test_connection_error` tests take a really long time, longer than what they should take given the timeout.

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -820,6 +820,8 @@ class FS(object):
 
         if not overwrite and self.exists(dst_path):
             raise errors.DestinationExists(dst_path)
+        if self.getinfo(src_path).is_dir:
+            raise errors.FileExpected(src_path)
         if self.getmeta().get('supports_rename', False):
             try:
                 src_sys_path = self.getsyspath(src_path)
--- a/fs/base.py
+++ b/fs/base.py
@@ -318,6 +318,8 @@ class FS(object):
         with self._lock:
             if not create and not self.exists(dst_path):
                 raise errors.ResourceNotFound(dst_path)
+            if not self.getinfo(src_path).is_dir:
+                raise errors.DirectoryExpected(src_path)
             copy.copy_dir(
                 self,
                 src_path,
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -99,6 +99,9 @@ class FTPFile(io.IOBase):
         self._read_conn = None
         self._write_conn = None
 
+    def __length_hint__(self):
+        return self.fs.getsize(self.path)
+
     def _open_ftp(self):
         ftp = self.fs._open_ftp()
         ftp.voidcmd(_encode('TYPE I'))
@@ -154,7 +157,7 @@ class FTPFile(io.IOBase):
     def readable(self):
         return self.mode.reading
 
-    def read(self, size=None):
+    def read(self, size=-1):
         if not self.mode.reading:
             raise IOError('File not open for reading')
 
@@ -163,8 +166,8 @@ class FTPFile(io.IOBase):
 
         conn = self.read_conn
         with self._lock:
-            while remaining is None or remaining:
-                if remaining is None:
+            while remaining:
+                if remaining < 0:
                     read_size = DEFAULT_CHUNK_SIZE
                 else:
                     read_size = min(DEFAULT_CHUNK_SIZE, remaining)
@@ -176,8 +179,7 @@ class FTPFile(io.IOBase):
                     break
                 chunks.append(chunk)
                 self.pos += len(chunk)
-                if remaining is not None:
-                    remaining -= len(chunk)
+                remaining -= len(chunk)
         return b''.join(chunks)
 
     def readline(self, size=None):
@@ -220,17 +222,14 @@ class FTPFile(io.IOBase):
     def truncate(self, size=None):
         # Inefficient, but I don't know if truncate is possible with ftp
         with self._lock:
-            if size is None:
-                size = self.tell()
+            size = size or self.tell()
             with self.fs.openbin(self.path) as f:
                 data = f.read(size)
             with self.fs.openbin(self.path, 'w') as f:
                 f.write(data)
                 if len(data) < size:
                     f.write(b'\0' * (size - len(data)))
-
-        self.pos = size or self.tell()
-        return self.pos
+        return size
 
     def seekable(self):
         return True
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -212,6 +212,8 @@ class FTPFile(io.IOBase):
                 remaining_data -= sent_size
                 self.pos += sent_size
 
+        return data_pos
+
     def writelines(self, lines):
         self.write(b''.join(lines))
 
@@ -226,6 +228,9 @@ class FTPFile(io.IOBase):
                 f.write(data)
                 if len(data) < size:
                     f.write(b'\0' * (size - len(data)))
+
+        self.pos = size or self.tell()
+        return self.pos
 
     def seekable(self):
         return True
@@ -252,6 +257,7 @@ class FTPFile(io.IOBase):
             if self._write_conn:
                 self._write_conn.close()
                 self._write_conn = None
+        return self.tell()
 
 
 class FTPFS(FS):
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -111,7 +111,7 @@ class FTPFile(io.IOBase):
         return self.fs.getsize(self.path)
 
     def _open_ftp(self):
-        ftp = self.fs._open_ftp(self.fs.encoding)
+        ftp = self.fs._open_ftp(self.fs.ftp.encoding)
         ftp.voidcmd(str('TYPE I'))
         return ftp
 
@@ -119,7 +119,7 @@ class FTPFile(io.IOBase):
     def read_conn(self):
         if self._read_conn is None:
             self._read_conn = self.ftp.transfercmd(
-                str('RETR ') + _encode(self.path, self.fs.encoding),
+                str('RETR ') + _encode(self.path, self.ftp.encoding),
                 self.pos
             )
         return self._read_conn
@@ -129,11 +129,11 @@ class FTPFile(io.IOBase):
         if self._write_conn is None:
             if self.mode.appending:
                 self._write_conn = self.ftp.transfercmd(
-                    str('APPE ') + _encode(self.path, self.fs.encoding)
+                    str('APPE ') + _encode(self.path, self.ftp.encoding)
                 )
             else:
                 self._write_conn = self.ftp.transfercmd(
-                    str('STOR ') + _encode(self.path, self.fs.encoding),
+                    str('STOR ') + _encode(self.path, self.ftp.encoding),
                     self.pos
                 )
         return self._write_conn
@@ -331,7 +331,7 @@ class FTPFS(FS):
         return _ftp
 
     def _manage_ftp(self):
-        ftp = self._open_ftp(self.encoding)
+        ftp = self._open_ftp(self.ftp.encoding)
         return manage_ftp(ftp)
 
     @property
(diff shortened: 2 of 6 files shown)


willmcgugan:
Thanks, Martin.

althonos:
Hi @willmcgugan, I managed to fix everything, but unfortunately `nose` seems to have an issue in Python 2 although everything works when I use `python2 -m unittest discover -v`. So I think I introduced an issue somewhere that makes `nose` unhappy, but I can't find it 😠

althonos:
@willmcgugan : clear to merge on my side.

willmcgugan:
Great. Give me a day or two to pick through this. Apologies in advance for any pedantry.

althonos:
No problem, I'll try to answer quickly any question you may have.

fs/ftpfs.py
```diff
@@ -99,16 +107,19 @@ def __init__(self, ftpfs, path, mode):
...
         self._write_conn = None
 
+    def __length_hint__(self):
+        return self.fs.getsize(self.path)
+
     def _open_ftp(self):
-        ftp = self.fs._open_ftp()
-        ftp.voidcmd(_encode('TYPE I'))
+        ftp = self.fs._open_ftp(self.fs.encoding)
+        ftp.voidcmd(str('TYPE I'))
```
willmcgugan: So `str` would be bytes on Python 2.7 and text type on Py3? Which is what the ftp library expects?

fs/ftpfs.py
```diff
@@ -212,20 +222,22 @@ def write(self, data):
...
+
     def writelines(self, lines):
         self.write(b''.join(lines))
 
     def truncate(self, size=None):
         # Inefficient, but I don't know if truncate is possible with ftp
         with self._lock:
-            if size is None:
-                size = self.tell()
+            size = size or self.tell()
```
willmcgugan: I think this would break if you did `truncate(0)`. Would it not truncate at the current file position?

fs/ftpfs.py
```diff
@@ -331,22 +345,28 @@ def ftp_url(self):
     @property
     def ftp(self):
         if self._ftp is None:
-            self._ftp = self._open_ftp()
+            self._ftp = self._open_ftp(self.encoding)
             self._welcome = self._ftp.getwelcome()
         return self._ftp
 
     @property
+    def encoding(self):
```
willmcgugan: Would it not be possible to detect and store the encoding in the `ftp` property? Just wondering the lazy detection of encoding would result in excessive requests.

fs/ftpfs.py
```diff
@@ -331,22 +345,28 @@ def ftp_url(self):
...
+        else:
+            return self._ftp.encoding
+
+    @property
     def features(self):
         """Get features dict from FTP server."""
         if self._features is None:
             try:
-                response = self.ftp.sendcmd(_encode("FEAT"))
+                response = _decode(self.ftp.sendcmd("FEAT"), "ascii")
```
willmcgugan: Is there a possibility of a unicode decode error here at all? I see the original did something similar, but it might be worth handling encoding problems for paranoia's sake.

tests/test_ftpfs.py
```diff
@@ -111,82 +82,59 @@ def test_manager(self):
...
-        server.stdout.readline()
-
-        if _WINDOWS_PLATFORM:
-            # Don't know why this is necessary on Windows
-            time.sleep(0.1)
+    user='user'
+    pasw='1234'
+
+    @classmethod
+    def setUpClass(cls):
```
willmcgugan: This does seem simpler than launching numerous ftp servers! Is it faster?

fs/base.py
```diff
@@ -318,6 +318,8 @@ def copydir(self, src_path, dst_path, create=False):
         with self._lock:
             if not create and not self.exists(dst_path):
                 raise errors.ResourceNotFound(dst_path)
+            if not self.getinfo(src_path).is_dir:
```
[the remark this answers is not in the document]
althonos: Calling `getinfo` will raise a `ResourceNotFound` as expected, whereas if you were to use `isdir` you would need one call to check for the existence of a resource, and a second call to check that the resource is a directory.

Since the base implementation uses `getinfo` as a base for `isdir`, `isfile` and `exist`, it will result in only one call to `getinfo` to do as such but two calls otherwise.

fs/ftpfs.py
```diff
@@ -15,7 +15,7 @@
 from ftplib import error_temp
 
 from six import PY2
-from six import text_type
+from six import text_type, binary_type
```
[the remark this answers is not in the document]
althonos: `six.binary_type` in Python 2 is `str`, not bytes :smile:

fs/ftpfs.py
```diff
@@ -99,16 +107,19 @@ def __init__(self, ftpfs, path, mode):
         self._read_conn = None
         self._write_conn = None
 
+    def __length_hint__(self):
```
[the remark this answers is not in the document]
althonos: Yes indeed, that totally went above my head when I made it !

fs/ftpfs.py
althonos: So, this is a summary from what I read during my researches:

`ftplib` expected binary (`str`) strings in Python 2, but was patched to simply decode text (`str`) strings in Python 3 and decode them as UTF-8. Because of this, there may be cases where a string is decode twice, or things like that. I'm not completely sure about it (the StackOverflow where people debated that stated the `ftplib` was a mess), but at least that provided good results with unicode path tests : when I did it another way, I'd start to have decoding/encoding errors in either Python 2 or Python 3.

fs/ftpfs.py
althonos: Yes you're right, I'll change it to `self.tell() if size is None else size`.

fs/ftpfs.py
althonos: I think so, but in both cases it only makes one additional call compared to the `master` implementation I guess. I could move it into the `ftp` property to encapsulate it.

fs/ftpfs.py
althonos: The [RFC](https://tools.ietf.org/html/rfc2389) does not say anything about encoding, but I believe features will only contain ASCII characters, so I guess there's no point over-interpreting the server response.

tests/test_ftpfs.py
althonos: Yes ! Running `tests.test_ftpfs` takes 6.5s on my machine :)

fs/ftpfs.py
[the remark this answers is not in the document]
willmcgugan: Yes, but on Python2.7

```python
>>> str is bytes
True
```

fs/ftpfs.py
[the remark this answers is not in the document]
willmcgugan: No worries. Don't think there is any shortcut for calculating number of lines for ftp files. Its an idea for MemoryFS though!

fs/ftpfs.py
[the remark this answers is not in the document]
althonos: I've been living in a lie for the last three years.

fs/ftpfs.py
willmcgugan: I think it might simplify things if the encoding detection logic is moved in to the `ftp`property. If the encoding defaulted to `latin-1` then you wouldn't need to pass it to `open_ftp`.
 
I'm also struggling a bit to understand why you need to open a new connection to detect the encoding. I think you should be able to call "FEAT" on first connection, and cache the encoding on the FTPFS. No need to make a new connection object.

fs/ftpfs.py
willmcgugan: I'd be tempted to handle the potential edge case where a response doesn't return 7-bit ascii. Maybe it won't ever happen, but there is a potential UnicodeDecodeError lurkng here.

How about `_decode` has unicode error handling set to `replace`? That way a flakey ftp server couldn't break this code.

fs/ftpfs.py
althonos: I agree with the first point, I'll change that.

For the second point: you have to set the encoding prior to connecting to the server. If you connect to the server with the default (latin-1), then change to utf-8 after seing the `UTF8` feature, the server will still think you're sending latin-1 encoded strings. So you to reset the server and use the proper encoding, you have to reconnect to the server with the new encoding.

tests/test_ftpfs.py
althonos: (except for `test_connection_error` which sometimes takes longer, sometimes not)

fs/ftpfs.py
althonos: Well, if a server can't even get the features list right, I'm sure the code will break elsewhere at some point 😁 But I'll change the code accordingly.

fs/ftpfs.py
willmcgugan: I think I've figured out something here. What was confusing me is that, I don't believe the encoding is ever sent to the ftp server. So the server is always going to send the same data regardless of what `FTP.encoding` is set to.

I've looked at the [source](https://github.com/python/cpython/blob/master/Lib/ftplib.py) and what appears to be happening is that its `ftplib` itself is using the encoding when it calls `self.sock.makefile` in the `connect` method. That creates a file-like object with the encoding when you connect.

So that would explain why you can't change the encoding after you have connected. But only due to a quite artificial limitation in ftplib. Otherwise you could just do a FEAT, and then set the encoding in the same connection.

It's really unfortunate.

I can think of a fudge, but its not very nice.

If the encoding is 'utf-8', you could re-*encode* the paths as 'latin-1' (effectively getting back the same bytes you received), then *decode* them as 'utf-8'. That should give you properly decode paths.

Another possibility is to do this if you detect 'utf-8' encoding:

```
ftp.file = ftp.sock.makefile('r', encoding='utf-8')
```

Both are hacks, so I'm not sure they are worth it. Otherwise we may be stuck with the single additional request to get the encoding...

althonos:
Any other change you want me to do ?

tests/test_ftpfs.py
willmcgugan: Is that a socket error? Maybe `socket.settimeout`

willmcgugan:
Looks good. Will have a quick pass later and merge.

Curious re your thoughts on https://github.com/PyFilesystem/pyfilesystem2/pull/68#discussion-diff-137926228R353 Looks like we don't strictly need 2 connections, to determine the encoding but the `ftplib` api doesn't allow for it.

willmcgugan:
Hmm. A few test fails on my Mac, but TravisCI isn't complaining.

althonos:
everything worked on my ArchLinux 😕 
Thanks for the merge ! Now's time for me to work the extensions compatible 😉

willmcgugan:
I'll document the test fails on the issues. Could use your feed back.

PR #72, opened 2017-09-06
Support FTP-URLs for proxies

muffl0n:
URLs for FTP over a proxy are in this format: ftp://user@ftp.example.org:password@proxy
This PR changes the Regex in "Registry" so that @ in usernames are allowed.

Compare: https://regex101.com/r/mPj5xu/1 vs https://regex101.com/r/mPj5xu/2


willmcgugan:
Is there precedence for doing that?

Because I'm planning to add a query string like syntax for extra information in FS URLs. For example:

```
ftp://user:password@example.org?proxy=ftp.proxy.org
```

muffl0n:
When I'm using lftp to connect via proxy this is the debug output:
```
-> % lftp user@ftp.example.org@proxy
Password: 
lftp user@ftp.example.org@proxy:~> debug
lftp user@ftp.example.org@proxy:~> ls
---- Connecting to proxy (x.x.x.x) port 21
<--- 220 proxy FTP server (Version x.x.x.x) ready.
---> FEAT
<--- 500 'FEAT': command unrecognized.
---> AUTH TLS
<--- 500 'AUTH': command unrecognized.
---> USER user@ftp.example.org
<--- 331 (not authenticated): Enter server password
---> PASS XXXX
...
```
So "user@ftp.example.org" is used as the username on the proxy. I found no RFC that does define this format as a standard though.

willmcgugan:
Yeah, I guess its a convention for some software, but not a standard per se.

I'd be inclined to not merge this, and ultimately add the query string support -- its a more general mechanism.

muffl0n:
Would work for me. :+1:

muffl0n:
After some digging around in Python's ftplib I still feel like my first approach would be the clearest way to go. If you take a look into the code you see no specific "proxy" settings. So you would still use "user@ftp.example.org" as the value for "user" and "ftp.proxy.org" as the server to connect to.

Sure, using "ftp://user@ftp.example.org?proxy=ftp.proxy.org" is more explicit but would also need some more parsing than using "ftp://user@ftp.example.org@ftp.proxy.org" which is well known from other tools.
See lftp's code as comparison: https://github.com/lavv17/lftp/blob/19b79494956dd26c44c296845b4900396c5c1e50/src/url.cc#L136
```
   // try to extract user name/password
   base=connect;
   scan=strrchr(base,'@');
```
The URL is scanned for the last occurence of @, allowing @ in the username for the server/proxy.

althonos:
I'm with Will on this one:
1. The proxied URL is quite confusing when you don't know it (what is the order ? Etc.)
2. It would be a lot more useful to have URL parameters, in this particular case but also for other filesystems

muffl0n:
# You would need a documentation either way. And why not use the format other popular programs use?
# I can totally understand this one. It would be far more consistent when you think about options for other filesystems.

Do you need a hand with the implementation? As I already mentioned this is a showstopper for us at the moment and I would give my left arm for a solution. :)

willmcgugan:
@muffl0n I'm afraid I'm still leaning towards the url parameters mechanism. If we go the proxy in username route, we would have a special case for FTP -- which I'm not keen on.

With regards to implementation, I'm not sure when I'll get to it. Possibly this weekend, but no guarantees. I don't see it being a massive undertaking. If the regex splits off the `?` character, then I think I can use the standard library to decode the rest.

althonos:
@willmcgugan : isn't there a `urllib.parse` function that handles parameters without even having to handle regexes manually ?

willmcgugan:
Yes, although it was actually [this syntax](http://commons.apache.org/proper/commons-vfs/filesystems.html) that I originally planned to implement, which isn't quite regular URL. 

urlparse also knows about existing protocols, and parses them accordingly. But we want FS URLs to be parsed in the same way regardless of the protocol.

willmcgugan:
@muffl0n I've added param support to a FS URLs.

The parse result object now has an extra value, `params`, which is a dictionary containing any query parameters.

Note, that the `@` symbol is problematic, so you may have to encode the proxy username as follows:

```
ftp://ftp.example.org?proxy=ftp.proxy.org&proxyuser=dave
```

Which would give you the following dict:

```json
{
    "proxy": "ftp.proxy.org",
    "proxyuser": "dave"
}
```

You can try this out with `fs==2.0.10a0`.

willmcgugan:
Going to close this PR. But feel free to carry on the discussion on https://github.com/PyFilesystem/pyfilesystem2/pull/75

Zujiry:
One question for me still stands: What if it is not the user for the proxy but the username for the actual ftp server that has an @ in it?

PR #75, opened 2017-09-09
added param support

willmcgugan:
Adds query string params to FS URLs (work in progress).


muffl0n:
Looks very promising to me! If you need any help just let me know! 
Thank you!

muffl0n:
I would like to add this test to emphasize the usage of user and password for the target ftp server
```
    def test_parse_user_password_proxy(self):
        parsed = opener.parse('ftp://user:password@ftp.example.org?proxy=ftp.proxy.org')
        expected = opener.registry.ParseResult(
            'ftp',
            'user',
            'password',
            'ftp.example.org',
            {
                'proxy': 'ftp.proxy.org'
            },
            None
        )
        self.assertEqual(expected, parsed)
```

willmcgugan:
@muffl0n @althonos This is in PyPi as `fs==1.0.10a2` if you would like to test.

Plan to do a full release tomorrow, if there are no issues.

fs/opener/parse.py
```diff
@@ -0,0 +1,105 @@
+"""
```
willmcgugan: This is just factored out, no material changes.

Issue #82, opened 2017-09-28
Error when Server (proxy) does not support "FEAT"-command


muffl0n:
Based on the new version 2.0.11 I did a quick patching of the fs.opener.ftpfs.FTPOpener#open_fs method to use our proxy if specified:
```
        if parse_result.params['proxy']:
            username = '{}@{}'.format(parse_result.username, ftp_host)
            ftp_host = parse_result.params['proxy']
        else:
            username = parse_result.username
        ftp_fs = FTPFS(
            ftp_host,
            port=ftp_port,
            user=username,
            passwd=parse_result.password,
        )
```

Unfortunately our proxy does not support the "FEAT"-Command and so I'm getting this error message:
```
Traceback (most recent call last):
  File "/foo/venv/lib64/python3.4/site-packages/fs/ftpfs.py", line 42, in ftp_errors
    yield
  File "/foo/venv/lib64/python3.4/site-packages/fs/ftpfs.py", line 372, in _read_dir
    self.ftp.retrlines(
  File "/foo/venv/lib64/python3.4/site-packages/fs/ftpfs.py", line 346, in ftp
    encoding = 'utf-8' if 'UTF8' in _ftp.sendcmd('FEAT') else 'latin-1'
  File "/usr/lib64/python3.4/ftplib.py", line 272, in sendcmd
    return self.getresp()
  File "/usr/lib64/python3.4/ftplib.py", line 245, in getresp
    raise error_perm(resp)
ftplib.error_perm: 500 'FEAT': command unrecognized.
```
For comparison, this is the output of lftp for the same connection:
```
-> % lftp user@ftp.example.org@proxy
Password: 
lftp user@ftp.example.org@proxy:~> debug
lftp user@ftp.example.org@proxy:~> ls
---- Connecting to proxy (x.x.x.x) port 21
<--- 220 proxy FTP server (Version x.x.x.x) ready.
---> FEAT
<--- 500 'FEAT': command unrecognized.
---> AUTH TLS
<--- 500 'AUTH': command unrecognized.
---> USER user@ftp.example.org
<--- 331 (not authenticated): Enter server password
---> PASS XXXX
...
```
See #72

I would expect PyFilesystem to not fail if "FEAT" is not supported.

willmcgugan:
Looks like an easy fix.

willmcgugan:
[...]

PR #84, opened 2017-10-04
Extend parsing for ftp directory listing

muffl0n:
I discovered that some directories in listings returned by a server running "ProFTPD 1.3.4a" were not listed with fs.base.FS#listdir.

E.g.
```
drwxr-xr-x   8 foo      bar          4096 Oct  4 09:05 test
```

This PR extends the regex and adds an appropriate test.


fs/_ftp_parse.py
```diff
@@ -29,7 +29,7 @@
     \s+?
     (\d+)
     \s+?
-    (\w{3}\s\d{2}\s+[\w:]+)
+    (\w{3}\s+\d+\s+[\w:]+)
```
willmcgugan: So this is the day number? Maybe the regex should be tighter and allow 1 or 2 digits?

muffl0n:
You are right, I fixed this one.

Btw, this seems to be a side effect of #82. As we know, when connecting via proxy the FEAT command fails and therefore the directory listings are not made via MLST, resulting in the behaviour seen here.

willmcgugan:
Thanks

PR #91, opened 2017-10-10
better handling of ftp features

willmcgugan:
Work in Progress

Handle utf8 encoding for ftp servers without opening another connection.

Also ensures that features are requested after login.

### DIFF
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -101,6 +101,7 @@ else:
 class FTPFile(io.IOBase):
 
     def __init__(self, ftpfs, path, mode):
+        super(FTPFile, self).__init__()
         self.fs = ftpfs
         self.path = path
         self.mode = Mode(mode)
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -10,7 +10,6 @@ import threading
 
 from contextlib import contextmanager
 from ftplib import FTP
-from ftplib import error_reply
 from ftplib import error_perm
 from ftplib import error_temp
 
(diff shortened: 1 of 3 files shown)


fs/ftpfs.py
```diff
@@ -317,17 +317,39 @@ def __str__(self):
...
+                        if line.startswith(' '):
+                            k, _, v = line[1:].partition(' ')
+                            self._features[k] = v
+                self.encoding = (
+                    'utf-8'
+                    if 'UTF8' in self._features
+                    else 'latin-1'
+                )
+                if not PY2:
+                    _ftp.file = _ftp.sock.makefile(
```
willmcgugan: Not sure how much of a hack this is. The `file` attribute of FTP objects is the only thing that references the encoding, by re-creating it, we can modify the encoding based on features.

althonos: It should be fine, I don't think the `ftplib` implementation will change since it is part of the stdlib.

Isn't your linter complaining about the seemingly-useless `self.ftp` lines in other methods though ?

willmcgugan: > It should be fine, I don't think the ftplib implementation will change since it is part of the stdlib.

That's my thinking.

> Isn't your linter complaining about the seemingly-useless self.ftp lines in other methods though ?

It is, and a few more complaints besides. I'll see if I can clear some of those up.

muffl0n:
I just tested these changes with our proxy and now I hate our proxy even more. :)
The code clearly does what it should by calling "FEAT" **after** logging in:
```
*resp* '220 proxy FTP server (Version x.x.x.x) ready'
*cmd* 'USER user@ftp.example.org'
*resp* '331 (not authenticated): Enter server password'
*cmd* 'PASS *********'
*resp* '230-Connected to server. Logging in...\n230-220 ProFTPD 1.3.4a Server (Debian) [x.x.x.x]\n230-331 Password required for user\n230 230 User user logged in'
*cmd* 'FEAT'
```
But that does not help with my use case as the proxy still does not proxy the command to the target server:
```
*resp* "500 'FEAT': command unrecognized."
```

I'm still trying to figure out how ncftp or lftp handle this.

PR #95, opened 2017-10-13
Relax six dependency

roll:
The `six` dependency prevents other software from work - https://github.com/Metatab/metapack/issues/4#issuecomment-336496954

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ REQUIREMENTS = [
     "appdirs~=1.4.3",
     "pytz",
     "setuptools",
-    "six~=1.10",
+    "six>=1.10,<2.0",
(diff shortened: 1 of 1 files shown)


althonos:
This is pointless since `six ~= 1.10` already means `six >= 1.10, <2`.

The example you are giving has a requirement which is `six ~= 1.10.0`, which means `six >= 1.10.0, <1.11`.

The latest version of `fs` will work with any `six` version superior to `1.10`, I just checked.

willmcgugan:
Closing this for now, as it doesn't look like it is the right fix.

Although if the ~= does turn out to be a problem, I could easily be persuaded to use the older syntax.

Issue #98, opened 2017-10-27
Support seek on the files opened from read-only zipfiles


rkhwaja:

willmcgugan:
I haven't confirmed this, but I suspect the [file objects](https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile.open) returned from `zipfile` just aren't seekable.

It's unfortunate but the only workaround would be to copy the file to a temp file first, which would incur a performance penalty.

Unless I can come up with a way of implementing seek that doesn't impact anything else, I may have to close this as 'wontfix'. Open to suggestions for a way to make that work...

althonos:
Some seek operations can be emulated for no cost. I'll explore this issue a
bit and come back with my findings.

althonos:
This is even better than what I expected : after studying the source of [`zipfile.py`](https://github.com/python/cpython/blob/master/Lib/zipfile.py), this is what I understood:

when opening a file in a zipfile, it will create a `_SharedFile` object that locks on reading, and does not support seeking, but maps the virtual position in the file to the actual position in the Zip file stream. So we could bypass the `_SharedFile` and implement locking ourselves in a custom `ZipExtFile`, which would be similar to what I did when I implemented `ISOFS`. That way, we could have real seeking as long as the underlying Zip file stream supports streaming.

willmcgugan:
Should be fixed in 2.0.15, thanks to @althonos

Issue #60, opened 2017-07-26
write_zip() doesn't store permissions into ZIP archives 


mashabow:
```console
$ ipython
Python 3.6.1 (default, Jul 19 2017, 13:24:40)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from fs.zipfs import ZipFS

In [2]: with ZipFS('test.zip', write='True') as z:
   ...:     z.makedir('foo', 0o777)
   ...:     z.makedir('bar', 0o755)
   ...:     z.makedir('baz', 0o700)
   ...:

In [3]:
Do you really want to exit ([y]/n)? y

$ zipinfo test.zip
Archive:  test.zip   274 bytes   3 files
?rw-------  2.0 unx        0 b- stor 26-Jul-17 15:00 bar/
?rw-------  2.0 unx        0 b- stor 26-Jul-17 15:00 baz/
?rw-------  2.0 unx        0 b- stor 26-Jul-17 15:00 foo/
3 files, 0 bytes uncompressed, 0 bytes compressed:  0.0%
```

It seems we have to set `ZipInfo.external_attr` in [`write_zip()`](https://github.com/PyFilesystem/pyfilesystem2/blob/44260e742b42b0e124e9f87dd115a70c4c857fc7/fs/compress.py#L25).

willmcgugan:
To be honest, I didn't realise that Zip files stored permissions info. But since they do, I agree ZipFS should write it.

willmcgugan:
Should be fixed in 2.0.15 Thanks to @althonos

PR #100, opened 2017-10-28
Seek files and extract permissions from `ZipFS`

althonos:
## About #98

The objects returned by `ZipFile.open` do not support seeking, but since they use an internal buffer we can use that buffer to our advantage. So here's what I did:

* `_ZipExtFile.seek` will try to use the internal zip buffer as much as possible, by manually changing the position of the internal file object in the buffer. If it's not possible, then: 
    - if we're seeking backward, we will have to open the file one more time (like `FTPFile` does)
    - if we're seeking forward, we read and discard until we reach the desired position
* `_ZipExtFile.read` and `_ZipExtFile.read1` were patched to prevent the internal file object from discarding the read buffer on EOF. Without this fudge, we would have to reopen the file each time the EOF is reached, even for a `_ZipExtFile.seek(-1, Seek.end)`.
* `_ZipExtFile.tell` will return the position of the file.

While I was at it, I implemented #60 as well :smile: : 

## About #60 

* When reading a zip file, permissions are extracted (but nothing else, since Zip files do not store owner / group of the archive), **providing that the zip file was created on a Unix system**, and that there are some permissions to extract.
* When writing a zip file, permissions from the `"access"` namespace are written (as well as the directory bit to indicate whether the zip entry is a file or a directory). If there are no permissions to write, `zip_info.external_attr` is left as default (`0`).

I also reorganized `ReadZipFs.getinfo` to only return the required namespaces, meaning the call to `ZipFile.getinfo` will be skipped if possible, and then did the same thing for `ReadTarFs` for the archives filesystems to be more consistent.

## People this may interest
* @willmcgugan 😄 
* @rkhwaja  
* @mashabow

### DIFF
--- a/fs/compress.py
+++ b/fs/compress.py
@@ -18,7 +18,7 @@ import six
 from .enums import ResourceType
 from .path import relpath
 from .time import datetime_to_epoch
-from .errors import NoSysPath
+from .errors import NoSysPath, MissingInfoNamespace
 from .walk import Walker
 
 
@@ -51,7 +51,8 @@ def write_zip(src_fs,
     )
     walker = walker or Walker()
     with _zip:
-        gen_walk = walker.info(src_fs, namespaces=["details", "stat"])
+        gen_walk = walker.info(
+            src_fs, namespaces=["details", "stat", "access"])
         for path, info in gen_walk:
             # Zip names must be relative, directory names must end
             # with a slash.
@@ -76,7 +77,13 @@ def write_zip(src_fs,
                 )
             zip_info = zipfile.ZipInfo(zip_name, zip_time)
 
+            try:
+                zip_info.external_attr = info.permissions.mode << 16
+            except MissingInfoNamespace:
+                zip_info.external_attr = 0o600
+
             if info.is_dir:
+                zip_info.external_attr |= 0x10
                 # This is how to record directories with zipfile
                 _zip.writestr(zip_info, b'')
             else:
--- a/fs/compress.py
+++ b/fs/compress.py
@@ -80,7 +80,7 @@ def write_zip(src_fs,
             try:
                 zip_info.external_attr = info.permissions.mode << 16
             except MissingInfoNamespace:
-                zip_info.external_attr = 0o600
+                pass
 
             if info.is_dir:
                 zip_info.external_attr |= 0x10
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -18,6 +18,16 @@ from .path import dirname, normpath, relpath, basename
 from .wrapfs import WrapFS
 from .permissions import Permissions
 
+__all__ = ['TarFS', 'WriteTarFS', 'ReadTarFS']
+
+
+if six.PY2:
+    def _get_member_info(member, encoding):
+        return member.get_info(encoding, None)
+else:
+    def _get_member_info(member, encoding):
+        return member.get_info()
+
 
 class TarFS(WrapFS):
     """Read and write tar files.
@@ -208,7 +218,7 @@ class ReadTarFS(FS):
         namespaces = namespaces or ()
         raw_info = {}
 
-        if _path in '/':
+        if not _path:
             raw_info['basic'] = {
                 "name": "",
                 "is_dir": True,
@@ -244,9 +254,7 @@ class ReadTarFS(FS):
                     "user": member.uname,
                 }
             if "tar" in namespaces:
-                raw_info["tar"] = raw_tar_info = member.get_info(*(
-                    [self.encoding, None] if six.PY2 else []
-                ))
+                raw_info["tar"] = _get_member_info(member, self.encoding)
                 raw_info["tar"].update({
                     k.replace('is', 'is_'):getattr(member, k)()
                     for k in dir(member)
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -204,59 +204,55 @@ class ReadTarFS(FS):
         return "<TarFS '{}'>".format(self._file)
 
     def getinfo(self, path, namespaces=None):
-        self.check()
+        _path = relpath(self.validatepath(path))
         namespaces = namespaces or ()
-        path = relpath(normpath(path))
-        if not path:
-            raw_info = {
-                "basic":
-                {
-                    "name": "",
-                    "is_dir": True,
-                },
-                "details":
-                {
+        raw_info = {}
+
+        if _path in '/':
+            raw_info['basic'] = {
+                "name": "",
+                "is_dir": True,
+            }
+            if 'details' in namespaces:
+                raw_info['details'] = {
                     "type": int(ResourceType.directory)
                 }
-            }
+
         else:
             try:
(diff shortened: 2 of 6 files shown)


willmcgugan:
Excellent. :) I'll have a quick scan.

fs/compress.py
```diff
@@ -76,7 +77,13 @@ def write_zip(src_fs,
                 )
             zip_info = zipfile.ZipInfo(zip_name, zip_time)
 
+            try:
```
willmcgugan: Information for external_attr was hard to find!

fs/compress.py
```diff
@@ -76,7 +77,13 @@ def write_zip(src_fs,
...
+                zip_info.external_attr = info.permissions.mode << 16
+            except MissingInfoNamespace:
+                zip_info.external_attr = 0o600
```
willmcgugan: Does this not have to be shifted left by 16?

fs/tarfs.py
```diff
@@ -204,59 +204,55 @@ def __str__(self):
...
-                "basic":
-                {
-                    "name": "",
-                    "is_dir": True,
-                },
-                "details":
-                {
+        raw_info = {}
+
+        if _path in '/':
```
willmcgugan: If `_path` has gone through `relpath` then I think `_path` could never be "/".

It should still trip with the empty string, but perhaps `not path` or `path == ""` is more explicit?

fs/zipfs.py
```diff
@@ -4,24 +4,110 @@
...
+        self._pos = 0
+        super(_ZipExtFile, self).__init__(_zip.open(name), 'r', name)
+
+    def read(self, size=-1):
+        if self._pos >= self._end:
+            return b''
+        elif size is None or size < 0:
+            size = self._end - self._pos
+            # NB(@althonos): do NOT replace by self._f.read() !
+            buf = b''.join([self._f.read(size-1), self._f._readbuffer[-1:]])
```
willmcgugan: For the special case of 2 strings, I'm pretty sure a string concatenation will be more efficient here and perhaps a little clearer...

```python
buf = self._f.read(size-1) + self._f._readbuffer[-1:]
```

fs/tarfs.py
```diff
@@ -204,59 +204,55 @@ def __str__(self):
...
                     "group": member.gname,
                     "permissions": Permissions(mode=member.mode).dump(),
                     "uid": member.uid,
                     "user": member.uname,
-                },
-                "tar": raw_tar_info,
-            }
+                }
+            if "tar" in namespaces:
+                raw_info["tar"] = raw_tar_info = member.get_info(*(
```
willmcgugan: Purely subjective, but I had to read this line twice. How about this as an alternative?

```python
raw_info["tar"] = raw_tar_info = (
    member.get_info(self.encoding, None)
    if six.PY2 else
    member.get_info()
)
```

fs/zipfs.py
```diff
@@ -4,24 +4,110 @@
...
+            self._f._offset += 1
+        elif self._f._offset + size <= len(self._f._readbuffer):
+            buf = self._f._readbuffer[self._f._offset:size+self._f._offset]
+            self._f._offset += size
+        else:
+            buf = self._f.read1(size)
+        self._pos += len(buf)
+        return buf
+
+    def seek(self, offset, whence=Seek.set):
```
willmcgugan: Really cool you got this to work when its not supported by the stdlib. :+1:

fs/zipfs.py
```diff
@@ -207,63 +293,63 @@ def _directory(self):
...
+                            if (not k.startswith('_') and
+                                not callable(getattr(zip_info, k)))
+                        }
+                    if "access" in namespaces:
+                        # check the zip was created on UNIX to get permissions
+                        if zip_info.create_system == 3:
+                            mode = zip_info.external_attr >> 16 & 0xFFF
+                            raw_info["access"] = {
+                                "permissions": Permissions(
+                                    mode=mode or 0o600  # default permission
```
willmcgugan: If the permissions bits are 0 does that indicate default permissions in a zip file? Because I think 0 is a valid (if unusual) value for unix permissions.

fs/compress.py
althonos: I think it would be even better to leave it as 0, which is the actual default anyway.

fs/tarfs.py
althonos: At start I did only `_path = self.validatepath(path)`, so I had to check that way, but since after that the failing tests remembered me `TarFile.getmember` expected a relative path, I added relpath and forgot to remove it.

fs/tarfs.py
althonos: Indeed, I have no idea why I did that in the first place when I PR'ed TarFS 😄 
IMO the best thing to do would be to define an external `getinfo(member, encoding)` function depending on the Python version once, to avoid checking the Python version on every getinfo call.

fs/zipfs.py
althonos: Is it ? 
I'm scared of doing string concatenation with `+` since a bad personal experience, but I'll follow you on that.

fs/zipfs.py
althonos: This is still kinda hackish, and won't work that well when going back and forth on a file larger than 4096B. I'll add a disclaimer to `_ZipExtFile.seek` about that. But at least it will work OK most of the time.

fs/zipfs.py
althonos: I checked by doing this:
```python
>>> import zipfile
>>> with zipfile.ZipFile('test.zip', 'w') as z:
...    info = zipfile.ZipInfo('test.txt')
...    z.writestr(info, '')
```
and then 
```
$ zipinfo test.zip
Archive:  test.zip
Zip file size: 114 bytes, number of entries: 1
?rw-------  2.0 unx        0 b- stor 80-Jan-01 00:00 test.txt
1 file, 0 bytes uncompressed, 0 bytes compressed:  0.0%
```

So I think actually having external attr as `0` means that there are no permission defined, which will result in `0o600` being applied when extracting on Unix. So it may be a better idea to only add the `access/permissions` info if external attr is the permissions are defined, and let the user do what he wants on no permissions.

Permissions of zip objects are a mess, since zip file don't store the uid/gid of the file owners, so even with permissions mode they are rendered pretty useless.

fs/compress.py
althonos: I ended up reading the `zipfile.py` source indeed ^_^

althonos:
@willmcgugan : clear to merge.

fs/zipfs.py
willmcgugan: Fair enough.

I've given thought to implementing some kind of custom PyFilesystem archive format that could capture everything that an FS object can expose. But it sounds like a lot of work, and there are probably bigger fish to fry...

willmcgugan:
Thanks. Will do a release soon.

fs/zipfs.py
althonos: This could be interesting indeed !

PR #102, opened 2017-11-09
added max_depth

willmcgugan:
**Work in progress.**

Adds a max_depth parameter to fs.walk.

### DIFF
--- a/fs/path.py
+++ b/fs/path.py
@@ -32,6 +32,7 @@ __all__ = [
     "iteratepath",
     "join",
     "normpath",
+    "parts",
     "recursepath",
     "relativefrom",
     "relpath",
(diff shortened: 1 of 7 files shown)


fs/path.py
```diff
@@ -256,6 +257,29 @@ def combine(path1, path2):
     return "{}/{}".format(path1.rstrip('/'), path2.lstrip('/'))
 
 
+def parts(path):
```
willmcgugan: Equivalent to https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parts

willmcgugan:
@althonos any thoughts?

althonos:
Fine by me. 

One remark though: the user may expect the `max_depths` to work even if `check_scan_dir` is overriden, which will not be the case here. 

A workaround would be:
```python
class BaseWalker(object):
# ... #
def _check_scan_dir(self, fs, path, info, depth):
    return (self.max_depth is None or depth < self.max_depth) \
       and self.check_scan_dir(fs, path, info, depth)

def check_scan_dir(self, fs, path, info, depth)
    """Override me !
    """
    return True
```

That way, `check_scan_dir` would be easier to override. (This may be applied to the other `check` functions as well).

PR #110, opened 2017-11-14
Fix #108 and add regression tests

### LINKED ISSUE #108: ZipFS: improper unpacking  of  (XML) files
I have this sample code what will read a ZIP file containing a valid `sd.xml` XML file:

```

import fs.opener

zf, _ = fs.opener.open('zip://sample.zip', writeable=False)
print zf.listdir(u".")
data = zf.open(u'sd.xml', 'rb').read()
print data
```

bildschirmfoto 2017-11-13 um 19 25 32

The screenshot shows that the closing tag `<smashdocc` is corrupted instead of `<smashdoc>`.

The second screenshot

bildschirmfoto 2017-11-13 um 19 27 55

shows that the `sd.xml` XML file is correct.

The related `sample.zip` can be found here: https://public.zopyx.com/sample.zip

althonos:
@zopyx, @willmcgugan 

When reading from a `ZipExtFile`, the buffer is completely discarded if the requested size is larger than `zipfile.ZipExtFile.MIN_READ_SIZE` (default is 4096 bytes). So I added more cases, to make sure the last buffer is never discarded. When reading a large file, as much as possible will be read with a single `read` call, then the remaining `MIN_READ_SIZE-1` bytes will be read independantly so that the underlying stream.

I added a regression test to `TestReadZipFS` (that did not pass before the fixes to `zipfs.py`) as well as more tests to check the underlying stream buffer is never empty after a read operation.

### DIFF
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -35,16 +35,21 @@ class _ZipExtFile(RawWrapper):
     def read(self, size=-1):
         if self._pos >= self._end:
             return b''
-        elif size is None or size < 0:
+        if size is None or size < 0:
             size = self._end - self._pos
-            # NB(@althonos): do NOT replace by self._f.read() !
-            buf = self._f.read(size-1) + self._f._readbuffer[-1:]
-            self._f._offset += 1
-        elif self._f._offset + size <= len(self._f._readbuffer):
+        if self._f._offset + size <= len(self._f._readbuffer):
             buf = self._f._readbuffer[self._f._offset:size+self._f._offset]
             self._f._offset += size
-        else:
+        elif size < self._end - self._pos < self._f.MIN_READ_SIZE:
             buf = self._f.read(size)
+        elif size < self._f.MIN_READ_SIZE:
+            buf = self._f.read(size-1) + self._f._readbuffer[-1:]
+        else:
+            buf = bytearray()
+            size -= self._f.MIN_READ_SIZE
+            buf.extend(self._f.read(size))
+            buf.extend(self._f.read1(self._f.MIN_READ_SIZE-1))
+            buf.extend(self._f._readbuffer[-1:])
         self._pos += len(buf)
         return buf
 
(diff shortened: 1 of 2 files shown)


fs/zipfs.py
```diff
@@ -35,16 +35,21 @@ def __init__(self, fs, name):
...
+        if self._f._offset + size <= len(self._f._readbuffer):
             buf = self._f._readbuffer[self._f._offset:size+self._f._offset]
             self._f._offset += size
-        else:
+        elif size < self._end - self._pos < self._f.MIN_READ_SIZE:
             buf = self._f.read(size)
+        elif size < self._f.MIN_READ_SIZE:
+            buf = self._f.read(size-1) + self._f._readbuffer[-1:]
+        else:
+            buf = bytearray()
```
willmcgugan: Shouldn't this method return a `bytes` object? Could be wrong, but I don't believe that `bytearray.extend` is more efficient that just adding strings. You're probably better of doing a `b''.join( ... )` here.

fs/zipfs.py
```diff
@@ -35,16 +35,21 @@ def __init__(self, fs, name):
...
+            size -= self._f.MIN_READ_SIZE
+            buf.extend(self._f.read(size))
+            buf.extend(self._f.read1(self._f.MIN_READ_SIZE-1))
```
willmcgugan: Shouldn't this be `read` rather than `read1`?

tests/test_zipfs.py
```diff
@@ -51,6 +51,21 @@ def load_archive(self):
     def remove_archive(self):
         os.remove(self._temp_path)
 
+    def test_large_read(self):
+        """Regression test for #108
+        """
+        # create a large file
+        test_fs = open_fs('mem://')
+        test_fs.setbytes('test.bin', b'a'*50000)
```
willmcgugan: Would it be worth adding a test for a large `read()`? i.e. without arguments?

althonos:
@willmcgugan : I think you're right, this is not a good direction for the code to take. At first I thought the readbuffer thing would work fine, but this is turning into an implementation-specific mess using private underlying attributes.
I'll close this PR, and create a proper PR to revert back to a purely emulated seeking solution.

PR #111, opened 2017-11-15
Revert back to emulate `ZipExtFile.seek`

althonos:
Using the lower level interface is:

* highly implementation-specific
* harder to understand
* generating bugs

This PR reverts back to a purely emulated seeking feature, i.e. *seeking forward by reading*.

### DIFF
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -65,13 +65,10 @@ class _ZipExtFile(RawWrapper):
 
         Note:
             Zip compression does not support seeking, so the seeking
-            is emulated. The internal decompression buffer will be used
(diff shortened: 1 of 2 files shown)


willmcgugan:
Fair enough. Using the buffer was clever, but obviously complicated things considerably. I imagine the 99% use case would be to read the entire file sequentially anyway.

althonos:
Yes, the ZIP format does not support random access anyway, so in the end the best practice would be either to read it sequentially or to decompress the archive.

willmcgugan:
Thanks. I'll do a release shortly.

PR #114, opened 2017-11-29
Fix test_ftp_parse to pass in 2018

bmwiedemann:
the embedded UNIX epoch values assumed that tm_year is 2017


willmcgugan:
A very timely fix, thanks.

PR #117, opened 2017-12-12
Fixed isbase method

fresheed:
Child and parent pathes were swapped by mistake. Documentation example for isbase didn't work because of that


willmcgugan:
Thanks.

PR #121, opened 2017-12-20
Add support for non-unicode, bytes-only paths on Linux and *nix

pombredanne:
This is a fix for #120
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -604,8 +604,6 @@ class FS(object):
 
         Parameters:
             path (str): A path on the filesystem.
-            as_bytes (bool, optional): If `True`, return the path as bytes using
-                the native filesystem encoding. (defaults to `False`).
 
         Returns:
             str: the *system path* of the resource, if any, always as Unicode.
@@ -630,6 +628,14 @@ class FS(object):
             A filesystem may return a system path even if no
             resource is referenced by that path -- as long as it can
             be certain what that system path would be.
+
+        Note:
+            The returned value is always Unicode. Some filesystems 
+            such as on Linux only deal with bytes. If you need to use
+            a path outside of Python, or if you are running Python 2
+            on Linux, you can get a proper byte value from a Unicode
+            path using the os.fsencode function on Python 3 or its
+            Pythyon 2 backport available here.
 
         """
         raise errors.NoSysPath(path=path)
--- a/fs/base.py
+++ b/fs/base.py
@@ -599,14 +599,16 @@ class FS(object):
         size = self.getdetails(path).size
         return size
 
-    def getsyspath(self, path):
+    def getsyspath(self, path, as_bytes=False):
         """Get the *system path* of a resource.
 
         Parameters:
             path (str): A path on the filesystem.
+            as_bytes (bool, optional): If `True`, return the path as bytes using
+                the native filesystem encoding. (defaults to `False`).
 
         Returns:
-            str: the *system path* of the resource, if any.
+            str: the *system path* of the resource, if any, always as Unicode.
 
         Raises:
             fs.errors.NoSysPath: If there is no corresponding system path.
--- a/fs/base.py
+++ b/fs/base.py
@@ -599,7 +599,7 @@ class FS(object):
         size = self.getdetails(path).size
         return size
 
-    def getsyspath(self, path, as_bytes=False):
+    def getsyspath(self, path):
         """Get the *system path* of a resource.
 
         Parameters:
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -252,12 +252,18 @@ class OSFS(FS):
         }
         return link
 
+    def _get_validated_syspath(self, path):
+        """Return a validated, normalized and eventually encoded string or byte
+           path.
+        """
+        path = path and fsdecode(path) or path
+        _path = self.validatepath(path)
+        return self._to_sys_path(_path, as_bytes=_NIX_PY2)
+
     def getinfo(self, path, namespaces=None):
         self.check()
         namespaces = namespaces or ()
-        path = path and fsdecode(path) or path
-        _path = self.validatepath(path)
-        sys_path = self._to_sys_path(_path, as_bytes=_NIX_PY2)
+        sys_path = self._get_validated_syspath(path)
         _lstat = None
         with convert_os_errors('getinfo', path):
             _stat = os.stat(sys_path)
@@ -266,7 +272,7 @@ class OSFS(FS):
 
         info = {
             'basic': {
-                'name': basename(_path),
+                'name': fsdecode(basename(sys_path)),
                 'is_dir': stat.S_ISDIR(_stat.st_mode)
             }
         }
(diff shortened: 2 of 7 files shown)


fs/base.py
```diff
@@ -599,14 +599,16 @@ def getsize(self, path):
         size = self.getdetails(path).size
         return size
 
-    def getsyspath(self, path):
+    def getsyspath(self, path, as_bytes=False):
```
willmcgugan: I'm not overly keen on overloading `getsyspath` like this.

How about leaving `getsyspath` as returning unicode, and add a `getnativepath` which may return any format (presumably bytes for *nix).

pombredanne: @coveralls your call. I had started doing so but then I found it heavy: to have another method, another error, etc. And since your docstring for `getsyspath` states:

```
        A system path is one recognized by the OS, that may be used
        outside of PyFilesystem (in an application or a shell for
        example). This method will get the corresponding system path
        that would be referenced by ``path``.
```

... I find it really hard to have another method with different docstring... since on *nix the real system path is bytes.

willmcgugan: So what would happen if you were to get a syspath with `as_bytes=False` where the path couldn't be decoded as unicode?

pombredanne: `as_bytes=False` is the default so you still always get `unicode` out. With `as_bytes=True` you will always get `bytes` out using the filesystem encoding.  There is no case I can think of where there would be nay possible issue (unless if you have no filesystem encoding at all.... but that is the source of so many possible errors and is a gross misconfiguration of the OS that's it is not worth catering to)

pombredanne:
@willmcgugan I have no idea why coveralls complains when I look at the coverage reports.

fs/base.py
pombredanne: but again, your call if you think that would break some API. I think not since this is only additive with a default.

pombredanne:
Sounds like this is decently passing the tests now and is ready for your review.

fs/base.py
willmcgugan: Will need to give this one some thought!

pombredanne: Note also that strictly speaking the conversion to bytes on *nix is only required on Python2. Python3 handles this fine internally (and the os.path and related stdlib have been sprinkled with plenty of os.fsencode/fsdecode as needed) .... So I could make the test for both `*nix and py2` ... now the need to get bytes is still valid on Py2 and Py3 as a unicode fsdecoded path such as in #120 cannot be used in a *nix shell outside of Python unless it is fsencoded first.

pombredanne:
as a fun side note: with this patch you could then claim that you are the only path abstraction library that works on all OS and FS encodings on Py2 and Py3. There are none anywhere else. I searched hard.

fs/base.py
pombredanne: The latest commit only deals with bytes internally on Python 2 and *nix. Everywhere else, this is unicode throughout. We still need of course the ability to fsencode some unicode, but frankly this could be left out entirely of the `getsyspath` method. Instead we could update the doc and stating that `getsyspath` always return Unicode and when and why this could be a problem, and how to `fsencode` the received unicode for use outside if needed.

This could be a happy middle ground.

pombredanne:
@willmcgugan With the latest commits, there is no change in API needed at all, and the internal use of fsencoded bytes is limited to *nix and Py2.

fs/base.py
willmcgugan: @pombredanne Sounds good. But could you drop the `as_bytes` then? Calling this method with the new parameter will break with most of the other filesystems and all of the external filesystems.

I'm still favouring the `getnativepath` which may return bytes or unicode, but leaving `getsyspath` as returning (surrogate encoded) unicode. The default `getnativepath` could just return the same value as `getsyspath`

fs/osfs.py
```diff
@@ -261,17 +297,17 @@ def getinfo(self, path, namespaces=None):
...
+        sys_path = self._get_validated_syspath(path)
         with convert_os_errors('listdir', path, directory=True):
-            names = os.listdir(sys_path)
+            names = [fsdecode(f) for f in os.listdir(sys_path)]
         return names
 
     def makedir(self, path, permissions=None, recreate=False):
         self.check()
         mode = Permissions.get_mode(permissions)
+        path = path and fsdecode(path) or path
```
willmcgugan: There's a convention we use that you don't modify a `path` variable once its passed in. The reason is that when you raise an error containing the path, the error message will contain the same path that the developer called the method with.

Could you replace this pattern with `path = fsdecode(path) if path else path` ? It's more familiar for developers who came to Python recently.

pombredanne: I can use a _path instead alright. But the and/or is more pythonic than if/else in my book. I can admit though that it may not be obvious ;)

setup.py
```diff
@@ -37,7 +37,8 @@
     install_requires=REQUIREMENTS,
     extras_require={
         "scandir :python_version < '3.5'": ['scandir~=1.5'],
-        ":python_version < '3.4'": ['enum34~=1.1.6']
+        ":python_version < '3.4'": ['enum34~=1.1.6'],
+        ":python_version < '3.0'": ['backports.os'],
```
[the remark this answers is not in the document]
pombredanne: Fixed

requirements.txt
```diff
@@ -3,3 +3,4 @@ enum34==1.1.6 ; python_version < '3.4'
 pytz
 setuptools
 six==1.10.0
+backports.os
```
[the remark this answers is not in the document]
pombredanne: Fixed

pombredanne:
> Thanks. Hope you don't mind my pickiness.

I can be much more picky than this ;)

fs/osfs.py
pombredanne: Fixed

pombredanne:
@willmcgugan The latest should have all your comments/feedback applied.

pombredanne:
@willmcgugan Some thing went south on the Py3 tests. Let me review this

pombredanne:
I cleaned up the regression I had introduced. We should be all good now.

pombredanne:
Unless you have some string objections, I will build a loval wheel until y

fs/base.py
pombredanne: > I'm still favouring the getnativepath which may return bytes or unicode, but leaving getsyspath as returning (surrogate encoded) unicode. The default getnativepath could just return the same value as getsyspath

Do you want this at all? I thing this is no longer needed as the doc explains that getsystempath is always unicode and that some use cases may need fsencoding to be usable

pombredanne:
Hold on merging this. There are plenty of issues left to test and fix on *nix with non-decodable paths. I am playing with beautiful crashes so far.

willmcgugan:
@pombredanne I think I can live without the `getnativepath` for now. I'll consider adding it myself later.

No rush, will be AFK for a few days over xmas.

willmcgugan:
@pombredanne Did you resolve the crashes? Is it ready for review?

willmcgugan:
The code has moved on a bit. And I've attempted to tackle this issue in https://github.com/PyFilesystem/pyfilesystem2/pull/167

willmcgugan:
Closing in favour of https://github.com/PyFilesystem/pyfilesystem2/pull/167

Thanks for taking the lead on this one pombredanne

PR #127, opened 2017-12-28
added extract method

willmcgugan:
Adds an `extract` method to base, to help with https://github.com/PyFilesystem/s3fs/issues/18

It should allow 'downloading' of files without the need for a temp file. The default is a simple file copy loop.

Open to suggestions re naming. Is `extract` appropriate?

@althonos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unrealeased]
+
+### Added
+
+- fs.getfile function
+
 ## [2.0.17] - 2017-11-20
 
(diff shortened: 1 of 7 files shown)


althonos:
The code looks good (just had a quick glance), and the use case seems generic enough to je reused in some other filesystems.

About the name, it could be confusing: `tarfile` method `extract` is actually closer to an `open` method (return a read-only file handle). The closest method of the standard library is probably `copyfileobj` of the `shutil` module. So maybe `copyfileto` or `dumpto` could be candidates as well (using the `to` suffix to emphasize the FS is writing to a file handle).

willmcgugan:
Good point re the name. I've noticed a pattern that probably dictates the name. We have `setbytes`, `settext`, `setfile`, and `getbytes`, `gettext`. But no `getfile`.

I think this `extract` method is exactly what I would expect `getfile` to be. i.e. the logical inverse of `setfile`.

Issue #112, opened 2017-11-23
Optimization of walk.files()


genben:
Use case:

- S3FS file system
- Looping over files in S3 bucket
- Many (thousands) file are stored on the same directory level

Current implementation of Walker.walk.files() doesn't use benefits of iterator when many files are stored in the same directory. It actually loads all files into memory and returns it to application at once. With S3 file system it may take minutes before the application actually receives first item to process.

I propose the following changes:

1. Add the following method to `class Walker`

        def _walk_files(self, fs, path, namespaces=None):
            """Walk files using iterator.
    
               This method allows to scan efficiently directories with large amount of files
               as it doesn't collect files in a temporary list.
    
               Using a *breadth first* search 
            """
            path = abspath(normpath(path))
            queue = deque([path])
            push = queue.appendleft
            pop = queue.pop
    
            while queue:
                dir_path = pop()
                for info in self._scan(fs, dir_path, namespaces=namespaces):
                    if info.is_dir:
                        if self.check_open_dir(fs, info):
                            push(join(dir_path, info.name))
                    else:
                        if self.check_file(fs, info):
                            yield (dir_path, info)

2. Modify `WalkerBase.files()` method as:

        def files(self, fs, path='/'):
            """Walk a filesystem, yielding absolute paths to files.
      
            Arguments:
                fs (FS): A filesystem instance.
                path (str, optional): A path to a directory on the filesystem.
      
            Yields:
                str: absolute path to files on the filesystem found
                recursively within the given directory.
      
            """
            iter_walk = iter(self._walk_files(fs, path=path))
            for dirname, info in iter_walk:
                yield join(dirname, info.name)

This will allow looping of files in S3 bucket efficiently.

Note, above suggestion uses a hard-coded *breadth first* search algorithm. It may be modified to support *depth first* search as well.

willmcgugan:
This is something I would like to look at. I think there may be a refactor I could do so that all methods of the Walker class could benefit from generators over lists. Thanks for prompting me.

willmcgugan:
I've made some change in the iter-walk branch https://github.com/PyFilesystem/pyfilesystem2/tree/iter-walk which uses iterators for more efficient walking. Would you mind giving that a try?

genben:
This is great news! 
I will check it and give a feedback.

willmcgugan:
Should be fixed in v2.0.18

PR #134, opened 2018-01-25
Walk iterators

willmcgugan:
This modifies the walk code to use iterators over lists where possible.

This is so that the `info`, `files` and `dirs` methods don't need to load an entire directory worth before they start yielding results--which should also be more memory efficient.

See https://github.com/PyFilesystem/pyfilesystem2/issues/112

Also includes some optimisations for file copying which will benefit S3FS and potentially other network filesystems that implement `getfile`.

### DIFF
--- a/fs/walk.py
+++ b/fs/walk.py
@@ -7,8 +7,7 @@ searching etc. See :ref:`walking` for details.
 
 from __future__ import unicode_literals
 
-import itertools
-
+from collections import defaultdict
 from collections import deque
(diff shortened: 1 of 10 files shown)


fs/walk.py
```diff
@@ -26,86 +25,7 @@
 """
 
 
-class WalkerBase(object):
```
willmcgugan: Didn't need a base class. Folded in to Walker.

PR #136, opened 2018-02-05
add proxy support to ftp

willmcgugan:
Add ftp proxy support.

### DIFF
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -285,7 +285,7 @@ class FTPFS(FS):
         timeout (int, optional): Timeout for contacting server (in seconds,
             defaults to 10).
         port (int, optional): FTP port number (default 21).
-        proxy (str): An ftp proxy, or ``None`` (default) for no proxy.
+        proxy (str): An FTP proxy, or ``None`` (default) for no proxy.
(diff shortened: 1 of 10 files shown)


fs/ftpfs.py
```diff
@@ -328,6 +331,22 @@ def __str__(self):
...
+        )
+
+    @property
+    def host(self):
+        return (
+            self.proxy
+            if self.proxy is not None else
+            self._host
+        )
+
```
muffl0n: I prefer using shorter code like 
```
return self.proxy or self._host
```
But I also understand that you wanted to keep user() and host() alike. If that was the goal you wanted to achieve I would at least change one of
```
if self.proxy is None else
```
and
```
if self.proxy is not None else
```
in order to be consistent. Maybe change host() to this?
```
def host(self):
        return (
            self._host
            if self.proxy is None else
            self.proxy
        )
```

willmcgugan: > return self.proxy or self._host

In this case I want to be explicit regarding about using `None` rather than anything that evaluates to False.

> in order to be consistent. Maybe change host() to this?

:+1:

Issue #137, opened 2018-02-07
tar fs TypeError: paths must be unicode (not str)


ReimarBauer:
Hi,

currently with 2.0.18 running in this issue. TypeError: paths must be unicode (not str)
An older build 2.0.12 works

```
Python 2.7.11 |Continuum Analytics, Inc.| (default, Dec  6 2015, 18:08:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
```

```
import fs
fs_url = u'tar:///tmp/TAR/example.tar'
_fs = fs.open_fs(fs_url)
for item in sorted(_fs.listdir(u"")):
     print(_fs.isdir(item))
```
TypeError: paths must be unicode (not str)

[example.tar.gz](https://github.com/PyFilesystem/pyfilesystem2/files/1703015/example.tar.gz)

```
tar -t -f example.tar 
bar/
bar/info
bar/foo/
bar/foo/hallo
foo/
foo/info
info
```

willmcgugan:
Should be fixed in master.

willmcgugan:
Fixed in 2.0.19

PR #139, opened 2018-02-10
tarfs encoding fix

willmcgugan:
Fixes encoding issue on Py27 with TarFS (https://github.com/PyFilesystem/pyfilesystem2/issues/137)


willmcgugan:
@althonos You might want to look over this one. The `listdir` method was returning bytes rather than unicode. Mostly due to the usual encoding confusion in the tarfile module.

Issue #144, opened 2018-02-23
`CreateFailed` should wrap the underlying error if any


althonos:
Hi Will, 
while I was working on `fs.sshfs`, I noticed that since `fs.open_fs` is expected to always raise `CreateFailed` when something goes wrong, it's not really easy to report what really went wrong.

I think that the `CreateFailed` exception should probably expose a `exc` attribute, like `OperationFailed` does. This way, it would be easier to manage the failed creation, depending on whether it comes from a `PermissionError`, a `TimeoutError`, a `ResourceNotFound` when `create=False`, etc. 

What do you think about this ?

willmcgugan:
:+1:

ReimarBauer:
Also raising that "universal error" gives the impression that we always need write access, which is wrong.

althonos:
Well, actually a `CreateFailed` exception is raised when *a FS object could not be instantiated*, not when *a filesystem could not be created*. Documenting this in the `CreateFailed` docstring is enough IMHO.

willmcgugan:
The name may be misleading. CreateFailed may be confused with the use of the word 'create' in the context of writing something.

We might want to think about a more descriptive name. Although we're stuck supporting CreateFailed for backwards compatibility.

althonos:
But we could create a new class (e.g. `InstantiationFailed`) and make `CreateFailed` simply an alias: then everywhere the error would actually be displayed as `InstantiationFailed`, etc.

willmcgugan:
Thanks @althonos for the fix.

PR #148, opened 2018-03-08
Allow wrapping exceptions in `CreateFailed`

althonos:
Hi Will,
I made it so `CreateFailed` accepts an additional `exc` argument, as discussed in #144.

I also added a decorator classmethod to `CreateFailed`, so that it is easier to wrap any errors occuring in a filesystem `__init__` method into a `CreateFailed` exception.


willmcgugan:
Thanks.

PR #154, opened 2018-03-21
Add typing comments

althonos:
Hi Will,

this is not really a PR for now, but just a notice that I'm working on adding typechecks to Pyfilesystem2.

I have tried them in a [side project of mine](https://github.com/althonos/InstaLooter) (which I also ported to use Pyfilesystem, mind you :wink:), and I think it should be interesting to have them for both finding bugs and for the end users to use type checkers as well. Since `fs` is actually quite strongly-typed this should not be *extremely* difficult (I hope I won't be deleting this sentence in a month).

I've added types to the `FS` base class, and just by doing that I was able to find the following bug (which is, IMO, the main perk of using them):

```
fs/appfs.py:48: error: Cannot assign to a method
```

*I'm also updating docstrings when I can, since I was a bit too excited to put *optional* everywhere; it should be used only to denote `None` is an acceptable value for an argument, not that a argument is a default argument !*

## Errors

* `fs.appfs`: overwriting `create` method with attribute in `AppFS.__init__` (**FIXED**)
* `WrapReadOnly.makedirs` arguments are not in the same order as in `FS.makedirs` (**FIXED**)
* `Info.__eq__` only works with objects that have a `raw` attribute (**FIXED**)
* `FTPFS.create` should return something, but doesn't (**FIXED**)
* `Info.size` is advertised to be always int, but can be `None` sometimes (**FIXED**, return type of `getsize` is `int`)
* `WrapFS` does not respect some `FS` signatures (**FIXED**)
* `WrapFS.filterdir` have wrong parameters order and type (**FIXED**)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,7 +90,8 @@ ENV/
 # Rope project settings
 .ropeproject
 
-
-#PyCharm
-
+# PyCharm
 .idea/
+
+# MyPy cache
+.mypy_cache
--- a/Makefile
+++ b/Makefile
@@ -24,3 +24,7 @@ testall:
 docs:
 	cd docs && make html
 	python -c "import os, webbrowser; webbrowser.open('file://' + os.path.abspath('./docs/build/html/index.html'))"
+
+.PHONY: typecheck
+typecheck:
+	mypy -p fs --ignore-missing-imports
--- a/Makefile
+++ b/Makefile
@@ -27,4 +27,4 @@ docs:
 
 .PHONY: typecheck
 typecheck:
-	mypy -p fs --ignore-missing-imports
+	mypy -p fs --config setup.cfg
--- a/README.rst
+++ b/README.rst
@@ -89,7 +89,7 @@ Credits
 -------
 
 -  `Will McGugan <https://github.com/willmcgugan>`__
--  `Martin Larralde <https://github.com/althonos>`__ for TarFS
+-  `Martin Larralde <https://github.com/althonos>`__
 -  `Giampaolo <https://github.com/gpcimino>`__ for ``copy_if_newer`` and
    ftp fixes.
 
--- a/fs/_repr.py
+++ b/fs/_repr.py
@@ -3,13 +3,14 @@
 
 from __future__ import unicode_literals
 
+import typing
 
-if False:
+if typing.TYPE_CHECKING:
     from typing import Text, Tuple
 
 
 def make_repr(class_name, *args, **kwargs):
-    # type: (Text, *object, **Tuple[object, object]) -> Text
+    # type: (Text, *object, **Tuple[Text, Text]) -> Text
     """Generate a repr string.
 
     Positional arguments should be the positional arguments used to
--- a/fs/_repr.py
+++ b/fs/_repr.py
@@ -3,7 +3,13 @@
 
 from __future__ import unicode_literals
 
+
+if False:
+    from typing import Text, Tuple
+
+
 def make_repr(class_name, *args, **kwargs):
+    # type: (Text, *object, **Tuple[object, object]) -> Text
     """Generate a repr string.
 
     Positional arguments should be the positional arguments used to
--- a/fs/_repr.py
+++ b/fs/_repr.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 
 import typing
 
-if typing.TYPE_CHECKING:
+if False:  # typing.TYPE_CHECKING
     from typing import Text, Tuple
 
 
--- a/fs/_repr.py
+++ b/fs/_repr.py
@@ -10,7 +10,7 @@ if typing.TYPE_CHECKING:
 
 
 def make_repr(class_name, *args, **kwargs):
-    # type: (Text, *object, **Tuple[Text, Text]) -> Text
+    # type: (Text, *object, **Tuple[object, object]) -> Text
     """Generate a repr string.
 
     Positional arguments should be the positional arguments used to
--- a/fs/appfs.py
+++ b/fs/appfs.py
@@ -9,13 +9,14 @@ subclasses of `~fs.osfs.OSFS`.
 
 # see http://technet.microsoft.com/en-us/library/cc766489(WS.10).aspx
 
+import typing
+
 from .osfs import OSFS
 from ._repr import make_repr
 from appdirs import AppDirs
 
-
-if False:  # typing imports
-    from typing import ClassVar, Optional, Text
+if typing.TYPE_CHECKING:
+    from typing import Optional, Text
 
 
 __all__ = ['UserDataFS',
@@ -51,7 +52,7 @@ class _AppFS(OSFS):
         )
 
     def __repr__(self):
-        return make_repr(
+        return make_repr(       # type: ignore
             self.__class__.__name__,
             self.app_dirs.appname,
             author=(self.app_dirs.appauthor, None),
@@ -65,6 +66,7 @@ class _AppFS(OSFS):
             self.__class__.__name__.lower(),
             self.app_dirs.appname
         )
+
 
 class UserDataFS(_AppFS):
     """A filesystem for per-user application data.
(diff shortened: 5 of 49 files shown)


willmcgugan:
:+1:

Great idea. I've been using mypy in the day job for a few weeks, and the benefits are very tangible.

althonos:
Quick report about how this is going:

The library is really easy to type. Excepting the IO, which seems to always be a pain.

I added `typing` as a requirement, because I thought it was nicer to use with some objects (like `Mode` being a `Container[Text]`). Where applicable, I used `typing.TYPE_CHECKING` to hide imports only used by the type checker.

I didn't change the API anywhere, except in `fs.multifs` where I turned `_required_writable` (which I renamed `_writable_required` to be more consistent with `_delegate_required`) into a method returning a filesystem instead of `None` to help typechecking:
```python
## before ##
self._writable_required()          # check write_fs is not None
self.write_fs.settext('...')       # we know this is not None, mypy doesn't !

## after ##
write_fs = self._writable_required()   # this always return a FS
write_fs.settext('...')
```

Where applicable, I added `FIXME` and `TODO` comments, so that this PR does not turn into a 10+ issue fixing mess but only adresses typing.

althonos:
Also, I have the following questions / suggestions after heavily reviewing the code:

* `fs.permissions`: `Permissions` should probably implement `__int__` 
* `fs.base`: `setbinfile`, `setfile` and `getfile` are really confusing: `getfile` and `setbinfile` require a binary file, `setfile` a text file, but has not `getfile` equivalent. I believe `getfile` should be renamed to `getbinfile`, and a text-file accepting `getfile` implemented (I know, this is backward incompatible).
* `fs.move`: `move_dir` could be implemented using a walker instead of `copy_dir`, to remove each file after it is moved instead of moving everything. This would be less storage-hungry, and easier to interrupt / resume transfers.
* `fs.multifs`: instead of using a namedtuple, we could implement a proper `_PrioritizedFS` class, and make it comparable. Why would we do that ?... so that we could use the `heapq` module and a priority queue instead of sorting the delegate filesystems over and over.

willmcgugan:
Glad the typing is going well. We can add mypy to Travis when its ready. 

> fs.permissions: shouldn't Permissions implement __int__ ?

To return `mode`? Good idea.

> fs.base: setbinfile, setfile and getfile are really confusing: getfile and setbinfile require a binary file, setfile a text file, but has not getfile equivalent. I believe getfile should be renamed to getbinfile, and a text-file accepting getfile implemented (I know, this is backward incompatible).

You may be right. I'd have to review that and refresh my memory. Backwards incompatible would not be a deal breaker, as long as we bump the major version number.

> fs.move: move_dir could be implemented using a walker instead of copy_dir, to remove each file after it is moved instead of moving everything. This would be less storage-hungry, and easier to interrupt / resume transfers.

A possible benefit of doing a copy + delete is that if it is interrupted for any reason, you still have all the original files. Since filesystems don't guarantee anything is written until closed, that might be a compelling reason to keep it this way. `shutil.move` (https://github.com/python/cpython/blob/master/Lib/shutil.py#L567) is implemented as a copy + delete of the whole dir, but I'm not sure if that was intentional or just the simplest implementation. Open for debate though...

> fs.multifs: instead of using a namedtuple, we could implement a proper _PrioritizedFS class, and make it comparable. Why would we do that ?... so that we could use the heapq module and a priority queue instead of sorting the delegate filesystems over and over

I think there is a mechanism in there to avoid repeated sorting. It should only sort when it first iterates over the filesystems, then the order will be cached.

althonos:
## Typing

So, I think I managed to type almost everything. Here are the quirks:

* We cannot guarantee that `FS.getsize` will only return `int`, as the `size` attribute may not be in the `details` namespace even if that namespace is present. So I made the `FS.getsize` return `Optional[int]`, but I'll add overrides so that most of its subclasses (`OSFS`, `MemoryFS`, etc.) return always `int`.

* I made `WrapFS` and `SubFS` generics: a `WrapFS[F]` is always guaranteed to have its `delegate_fs` return a `F` instance, where `F` is a `FS` subclass. I made them covariant, so that `WrapFS[UserCacheFS]` is considered as a `WrapFS[OSFS]`, for instance.

* Unfortunately, we cannot guarantee that the `opendir` method for any filesystem `F` will return a `SubFS[F]`: namely, `MountFS` and `MultiFS` may return any `SubFS`. As such, the `FS.opendir`, `FS.makedir` and `FS.makedirs` are declared to only return `SubFS[FS]`; specific signature was added to other subclasses (for instance, `FTPFS.opendir()` returns `SubFS[FTPFS]`).

* Typing file-like objects is a mess, because the `io` module stubs don't consider `RawIOBase` to be a `BinaryIO`, and it's a pain to make everything compatible with both `RawIOBase` and `BinaryIO`. So I just ended with having `FS.openbin` return `BinaryIO`, `FS.open` return `IO`, and ignore all the warnings that could occur.


## `move_dir`

I'll open a PR later, to implement another `move_dir` implementation in another function: this way, we keep the current behaviour, but offer the possibility to use another one.


## `MultiFS` sorting

You're right, it only sorts once, and sorts once again when a new FS is added. I was wrong with the heap, the right data structure to use would be a BST, but that does not make sense using them here, since we are iterating over the sorted filesystems more often than we are adding a new one. So the right structure is indeed an array in this case.

willmcgugan:
Nice. Give me a little time to go over this, but its a no-brainer as far as merging goes.

> We cannot guarantee that FS.getsize will only return int, as the size attribute may not be in the details namespace even if that namespace is present. So I made the FS.getsize return Optional[int], but I'll add overrides so that most of its subclasses (OSFS, MemoryFS, etc.) return always int.

The [info docs](https://docs.pyfilesystem.org/en/latest/info.html#details-namespace) mandate that a `size` key will always be available on the `details` namespace, so I think its fare to say that `FS.getsize` will always return an int. Although I haven't explicitly said in the docs that there will always be a `details` namespace, but I think that we can guarantee it.

But even so, we can just make the guarantee that `getsize` will always return an int, and any FS that doesn't is broken.

> I'll open a PR later, to implement another move_dir implementation in another function: this way, we keep the current behaviour, but offer the possibility to use another one.

:+1: Any thoughts on how to name that function?

althonos:
> The info docs mandate that a size key will always be available on the details namespace, so I think its fare to say that FS.getsize will always return an int. Although I haven't explicitly said in the docs that there will always be a details namespace, but I think that we can guarantee it. But even so, we can just make the guarantee that getsize will always return an int, and any FS that doesn't is broken.

No problem about a missing namespace: in this case, `getsize` will raise a `MissingNamespacError`, which is not conflicting with its return type. 

> Any thoughts on how to name that function?
I don't know, something like `move_dir_recursive` or `move_dir_incremental`, but I'm convinced a better name can be found.

althonos:
There's one thing left, I used `typing.TYPE_CHECKING` as a safeguard for typing imports, but this is only available in `typing > 3.5.1`, meaning `Python 3.5.1` will complain about this. So I need to change that or add `!= 3.5.1` to the `python-requires` setup key.

willmcgugan:
Ok, I grok it now. typing was added to the stdlib in 3.5, and there is a backport for < 3.5.

How about this monstrosity?

```
if getattr(typing, 'TYPE_CHECKING', True):
```

althonos:
> I'm a little puzzled. The unit tests passed in Python2.7+ Does that mean that TYPE_CHECKING is missing from just 3.5.1
Backports are available only for Python < 3.5, so you can't use `typing == 3.6` with Python 3.5 unfortunately...

> How about this monstrosity? `if getattr(typing, 'TYPE_CHECKING', True)`
I think it would be clearer to stay with the following in that case (and no added run-time cost):
```python
if False: # typing.TYPE_CHECKING
    from typing import ...
```

willmcgugan:
Thanks @althonos 

Coverage has dropped, but I see its just those typing imports. I'll exclude those.

PR #161, opened 2018-04-16
Fix issue #160 (TarFS implicit directory bug)

### LINKED ISSUE #160: PyFilesystem can't find existing file in tar-fs
I wrote a stackoverflow post about this, I'm not entirely sure, but I suspect this to be a bug:

https://stackoverflow.com/questions/49811622/pyfilesystem-cant-find-existing-file-in-tar-fs

althonos:
Hi Will, 

just a quick PR to fix #160, with regression tests. Is there any `unique` iterator filter in the `fs` codebase ? It could avoid using a set just to filter out duplicates in the `listdir` patch.

### DIFF
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -16,7 +16,7 @@ from .enums import ResourceType
 from .info import Info
 from .iotools import RawWrapper
 from .opener import open_fs
-from .path import dirname, normpath, relpath, basename
+from .path import dirname, relpath, basename, isbase, parts, frombase
 from .wrapfs import WrapFS
 from .permissions import Permissions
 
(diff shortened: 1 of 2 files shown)


willmcgugan:
Great, thanks. Will do a release once that's in.

multifs filters out duplicate paths, and I did some research. It seems that this is the fastest way to filter duplicates while preserving order:

```
directory = list(OrderedDict.fromkeys(directory))
```

althonos:
> multifs filters out duplicate paths, and I did some research. It seems that this is the fastest way to filter duplicates while preserving order: `directory = list(OrderedDict.fromkeys(directory))`

I use the `unique_everseen` recipe most of the time, but this is a nice trick ! I'll add it right now.

althonos:
Yes, no more work intended !

 I'll rebase the typing branch as well.

willmcgugan:
Great, thanks!

PR #164, opened 2018-04-29
Tox fixes

willmcgugan:
@althonos This started life as a few tweaks to get the coverage up. I found some Python 3.5.1 issues which I had to work around:

 - typing.Text is not available
 - typing.overload decorator only works in stubs

The typing.Text was easy to work around, but I had to sacrifice the typing.overload.

Let me know if you think I made the right call. Will do a release soon.


fs/opener/ftpfs.py
```diff
@@ -43,8 +43,8 @@ def open_fs(self,
...
-
-        if dir_path:
-            return ftp_fs.opendir(dir_path, factory=ClosingSubFS)
-
-        return ftp_fs
+        return (
+            ftp_fs.opendir(dir_path, factory=ClosingSubFS)
+            if dir_path else
+            ftp_fs
+        )
```
althonos: Changing this has my `mypy` (v0.590) raise 
```
fs/opener/ftpfs.py:46: error: Incompatible return value type (got "FS", expected "Union[FTPFS, SubFS[FTPFS]]")
```
whereas it did not before (that's why I changed it).

Error does not occur with:
```python
if dir_path:
    return ftp_fs.opendir(dir_path, factory=ClosingSubFS)
else:
    return ftp_fs
```

fs/info.py
```diff
@@ -58,38 +58,13 @@ def __eq__(self, other):
...
-    @typing.overload
-    def get(self, namespace, key):
-        # type: (Text, Text) -> Optional[T]
-        pass
-
-    @typing.overload
-    def get(self, namespace, key, default):
-        # type: (Text, Text, T) -> T
-        pass
-
```
althonos: Just as there is a `_fscompat` module, maybe we could have a `_tycompat` where `overload` is defined as a no-op decorator when it cannot be imported from the `typing` module ? 

This is part of the public API, and I think it's a pity to remove a finer typed API just because of a single Python version.

fs/opener/ftpfs.py
willmcgugan: Ah right. I had assumed wrongly that mypy would consider them to be equivalent.

fs/ftpfs.py
```diff
@@ -96,16 +96,14 @@ def manage_ftp(ftp):
...
         try:
             ftp.quit()
-        except:  # pragma: nocover
+        except:  # pragma: no cover
             pass
 
 
 def _parse_ftp_error(error):
-    # type: (ftplib.Error) -> Tuple[Union[int, Text], Text]
+    # type: (ftplib.Error) -> Tuple[Text, Text]
```
willmcgugan: The code doesn't actually need to be an integer. We only ever test for equality. Simplifies things a little.

willmcgugan:
@althonos Second stab. Thoughts?

PR #167, opened 2018-05-02
handle broken unicode paths

willmcgugan:
# Work in progress

Attempt to tackle broken unicode filenames on Linux 

FIxes https://github.com/PyFilesystem/pyfilesystem2/issues/120

### DIFF
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -92,10 +92,13 @@ class OSFS(FS):
         """Create an OSFS instance.
         """
         super(OSFS, self).__init__()
-        root_path = fsdecode(fspath(root_path))
+        if isinstance(root_path, bytes):
+            root_path = fsdecode(root_path)
+        self.root_path = root_path
+        _root_path = fsdecode(fspath(root_path))
         _root_path = os.path.expanduser(os.path.expandvars(root_path))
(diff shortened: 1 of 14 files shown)


fs/osfs.py
```diff
@@ -368,7 +374,10 @@ def opendir(self, path, factory=None):
 
     def getsyspath(self, path):
         # type: (Text) -> Text
-        sys_path = self._to_sys_path(path)
+        sys_path = os.path.join(
```
willmcgugan: This method still returns unicode, so no change in behaviour.

If you want to call an `os` method on python2.7 and support broken unicode you will need to process the returned unicode path with `fsencode`. On Python3 fsencode is not required (but won't break).

That would be a source of annoyance. So I'm considering a `getospath` method which `fsencode`s the syspath.

@pombredanne @althonos  Thoughts?

tests/test_encoding.py
```diff
@@ -0,0 +1,55 @@
...
+import unittest
+
+import six
+
+import fs
+from fs.osfs import OSFS
+
+
+
+@unittest.skipIf(platform.system() == 'Darwin', "Bad unicode not possible on OSX")
```
willmcgugan: OSX raise IOErrors when trying to create broken unicode paths.

PR #172, opened 2018-05-10
Use `Markdown` readme and add project urls

althonos:
PyPI is gone, long live PyPI ! The new PyPI (known as *warehouse* before it was released) introduces new features, such as:

* Markdown support for READMEs
* Project URLs to be displayed on the PyPI page directly

I did a test deployment at https://test.pypi.org/project/fs/ and it looks just fine with the Markdown readme !


willmcgugan:
:+1:

Did you use twine to do the upload? I recently followed [this guide](https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi) to do it for Lomond.

We might need up update the Makefile.

althonos:
Yes I did ! Although doing it myself was a first, I use Travis for deployment so I didn't see twine in action before (but I knew they changed their PyPI deployment provider to use it).

althonos:
(By the way, would you be interested in setting up a deployment pipeline ? For a pure Python library it's extremely easy)

willmcgugan:
Potentially. How would it work?

althonos:
For `fs.sshfs`, I configured it to release whenever I do a tagged commit : on a successful job for a tagged commit, it will build a source distribution and universal wheel and upload it to PyPI. So when I need to release, I bump the version in a commit I then tag, and push all.

willmcgugan:
That does sound convenient. TBH It never occurred to me to automate it. I just do `make release` then a tag. But I think I forget the tag from time to time, so automation may be the way to go.

willmcgugan:
Thanks for the PR.

PR #173, opened 2018-05-24
Enable `timeout` in URI

geoffjukes:
Enable `timeout` value in URI, with default matching documented default (10 seconds).


willmcgugan:
Thanks.

PR #176, opened 2018-06-20
Make filesystems pickleable

martindurant:
The presence of a thread lock in the filesystem makes is unpickleable. It can be recreated at each unpickling for each new instance, instead.
Note that all state is carried over from one instance to the other - but it should be up to the FS implementations to decide how to pickle themselves, not the base class.

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -94,6 +94,15 @@ class FS(object):
         """
         self.close()
 
+    def __getstate__(self):
+        out = self.__dict__.copy()
+        out.pop('_lock', None)
+        return out
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._lock = threading.RLock()
+
     @property
     def walk(self):
         # type: (_F) -> BoundWalker[_F]
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
+import pickle
 import unittest
 
 try:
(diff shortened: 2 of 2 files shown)


willmcgugan:
Picking is something that PyFilesystem had, but was intentionally not support in PyFilesystem2.

Pickling doesn't make sense for many filesystems. What should happen if you pickle a tempfs or a memoryfs for example? Should network filesystems store authentication? What happens if you unpickle an OSFS on another machine? There are too many edge cases to make it a general concept.

If you want to restore a filesystem in a replaceable way, look into [FS URLs](https://docs.pyfilesystem.org/en/latest/openers.html).

martindurant:
> Pickling doesn't make sense for many filesystems.

However, I don't think it is the business of the base class to, in effect, prevent pickling of all derived classes. It should be up to specific implementations to decide what they want to do.

> Should network filesystems store authentication?

[s3fs](http://s3fs.readthedocs.io/) is totally fine with pickling, and has well-defined behaviour for authentication. 

> What happens if you unpickle an OSFS on another machine

This is a real use case, you may share an NFS, for example. The unpickler could check for the existence of the path, but, again, this is up to the implementation.

This PR exists as a precursor to any possible use by [Dask](http://dask.pydata.org/en/latest/).

willmcgugan:
PyFilesystem guarantees that implementations are interchangeable. If behaviour is left up to the implementation then that guarantee is broken. You can no longer swap one for the other and expect it to work.

FS URLs are a better way of serialising filesystems. The behaviour is well defined, and it suffers none of the issues with picking.

Issue #183, opened 2018-07-06
FSErrors freeze Python when thrown in multiprocessing context 


Spacerat:
I am using:
- Python 3.6.4
- fs==2.0.24

I discovered this issue when trying to use `listdir` in the context of a concurrent.futures.ProcessPoolExecutor. When a child of `FSError` is thrown in a multiprocessing context, the program never exits.

Here is a minimal reproduction

```py
import concurrent.futures
import fs.errors

def inner():
    # A simple ValueError would be bubbled to the parent process correctly
    raise fs.errors.ResourceError("foo")

def main():
    with concurrent.futures.ProcessPoolExecutor() as pool:
        pool.submit(inner).result()
        # Program never exits.


if __name__ == "__main__":
    main()
```

The following error is displayed, and the program hangs:
```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/process.py", line 272, in _queue_management_worker
    result_item = reader.recv()
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/multiprocessing/connection.py", line 251, in recv
    return _ForkingPickler.loads(buf.getbuffer())
TypeError: __init__() missing 1 required positional argument: 'path'
```

I have narrowed the implementations of the pyfilesystem2 errors down to the minimum required to trigger the problem:

```py
class MyFSError(Exception):
    def __init__(self):
        # Calling __init__ as __init__(str(self)) here seems to fix the problem.
        # but it only works if all child exceptions have only a single argument.
        super(MyFSError, self).__init__() 


class MyResourceError(MyFSError):
    def __init__(self, path):
        # Supplying a default argument to MyResourceError also fixes the problem.
        super(MyResourceError, self).__init__()

def inner():
    raise MyResourceError("foo")
    # Note that `raise MyFSError()` does not cause the same problem
```

[...]

PR #184, opened 2018-07-08
Fix multiprocessing hang on FSErrors

Spacerat:
This is a fix for the problem of subclasses of FSError/Exception causing multiprocessing to hang ( issue #183 ) if they change the arguments to `__init__`.  I went with @willmcgugan 's suggestion to implement `__reduce__` in each subclass wherever it was required.

### DIFF
--- a/fs/errors.py
+++ b/fs/errors.py
@@ -59,10 +59,14 @@ class MissingInfoNamespace(AttributeError):
 
     def __init__(self, namespace):
         # type: (Text) -> None
+        self.namespace=namespace
         msg = "namespace '{}' is required for this attribute"
         super(MissingInfoNamespace, self).__init__(
             msg.format(namespace)
         )
+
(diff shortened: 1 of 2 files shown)


fs/errors.py
```diff
@@ -59,11 +59,15 @@ class MissingInfoNamespace(AttributeError):
 
     def __init__(self, namespace):
         # type: (Text) -> None
+        self.namespace=namespace
```
willmcgugan: Tiny pep8 thing, should be whitespace around `=`

willmcgugan:
@Spacerat Looks good, thanks. Just a tiny request in the comments, and its good to merge.

Spacerat:
:+1: thanks for the review @willmcgugan . I've fixed the formatting error now

willmcgugan:
Thanks.

PR #186, opened 2018-07-12
Fix `test_entry_point_create_error` (fix #185)

### LINKED ISSUE #185: Test failure under Python 3.7
```
======================================================================
ERROR: test_entry_point_type_error (tests.test_opener.TestRegistry)
----------------------------------------------------------------------
  File "/build/python-fs/src/pyfilesystem2-2.0.24/tests/test_opener.py", line 213, in test_entry_point_type_error
    opener.open_fs('test://')
  File "/build/python-fs/src/pyfilesystem2-2.0.24/fs/opener/registry.py", line 192, in open_fs
    default_protocol=default_protocol
  File "/build/python-fs/src/pyfilesystem2-2.0.24/fs/opener/registry.py", line 147, in open
    opener = self.get_opener(protocol)
  File "/build/python-fs/src/pyfilesystem2-2.0.24/fs/opener/registry.py", line 97, in get_opener
    if not isinstance(opener, object) or not issubclass(opener, Opener):
  File "/usr/lib/python3.7/abc.py", line 143, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class

----------------------------------------------------------------------
Ran 1350 tests in 7.726s
```

althonos:
Proposed fix for #185 :

Instead of monkey patching `issubclass` for it to accept `MagicMock` as an `Opener` subclass, I create a mock opener that raises a `ValueError` on instantiation. Less dirty, and works on Python 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,21 @@
+sudo: false
 language: python
 
-python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-
-os:
-  - linux
+matrix:
+  include:
+  - python: "2.7"
+    env: SETUPTOOLS=setuptools
+  - python: "3.3"
+    env: SETUPTOOLS=setuptools~=39.2
+  - python: "3.4"
+    env: SETUPTOOLS=setuptools
+  - python: "3.5"
+    env: SETUPTOOLS=setuptools
+  - python: "3.6"
+    env: SETUPTOOLS=setuptools
 
 before_install:
-  - pip install setuptools pip -U
+  - pip install $SETUPTOOLS pip -U
   - pip --version
   - pip install -r testrequirements.txt
   - pip freeze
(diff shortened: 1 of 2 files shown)


althonos:
NB: I also added a version constraint for `setuptools` in Python 3.3, since agressively updating will cause the build to fail (`setuptools 40.0` dropped support for Py3.3).

willmcgugan:
Thanks @althonos 

How would you feel if we dropped support for Py3.3 ?

althonos:
Depends. For now it does not cost anything to keep support besides installing an older setuptools, so I don't feel we should drop it. We could move it to the *allowed failures* for the time being though.

willmcgugan:
I think your right about supporting it so long as its not a major effort. 

Thanks for the PR. Is it ready to go?

willmcgugan:
Thanks

althonos:
Yes it was :wink:

Issue #182, opened 2018-07-04
Open files with O_EXCL in Python 2


sqwishy:
`io.open` supports taking file descriptor instead of a file path (and can take ownership of the fd). `os.open` lets you open a file in exclusive mode and returns a file descriptor. You could do something like the following to open the file in the mode specified.

```
diff --git a/fs/osfs.py b/fs/osfs.py
index e53e5e2..9b2c330 100644
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -422,8 +422,8 @@ class OSFS(FS):
         _path = self.validatepath(path)
         sys_path = self._to_sys_path(_path)
         with convert_os_errors('open', path):
-            if six.PY2 and _mode.exclusive and self.exists(path):
-                raise FileExists(path)
+            if six.PY2 and _mode.exclusive:
+                sys_path = os.open(sys_path, os.O_CREAT | os.O_EXCL)
             _encoding = encoding or 'utf-8'
             return io.open(
                 sys_path,
```

*(I tried running the tests to see what this breaks, but for some reason tox wouldn't stop barfing at me even on regular v2.0.24 without this change.)*

willmcgugan:
To avoid that extra call to `exists`? Seems like a good idea.

What was tox complaining about?

sqwishy:
It's mainly since exclusive mode is supposed to be atomic. Using `exists()` allows a race condition that O_EXCL shouldn't have.

>What was tox complaining about?

Literally everything under the moon. Maybe my PATH is messed up or my version of gcc is older than time itself; I'm sure it's my environment. I can look more closely later today if it would help.

sqwishy:
On a different machine, tox runs correctly, as far as I can tell, for 2.7 and 3.6 with four test failures in each version (`tests.test_copy.TestCopy`). It doesn't fail more when I apply the diff above.

The diff assumes that if you're in exclusive mode, you want to create. Which is a bit weird to me at first, but `open(2)` states
>In general, the behavior of O_EXCL is undefined if it is used without O_CREAT.  There is one exception: on Linux 2.6 and later, O_EXCL can be used without O_CREAT if pathname refers to a block device.  If the block device is in use by the system (e.g., mounted), open() fails with the error EBUSY.

[...]

PR #189, opened 2018-07-17
Use `os.open()` in osfs in exclusive mode (#182)

sqwishy:
`io.open()` takes a file descriptor, `os.open()` returns one and lets you pass O_EXCL to open the file.

### DIFF
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -422,8 +422,8 @@ class OSFS(FS):
         _path = self.validatepath(path)
         sys_path = self._to_sys_path(_path)
         with convert_os_errors('open', path):
-            if six.PY2 and _mode.exclusive and self.exists(path):
-                raise FileExists(path)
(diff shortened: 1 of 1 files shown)


willmcgugan:
Thanks, will get to this soon. Bit of a backlog...

willmcgugan:
Thanks

sqwishy:
@willmcgugan Hey, I noticed today that the changes in this pull request totally don't work. If you pass a file descriptor to `io.open()` and then either call `close()` or `__exit__()` (exit the context manager) on the file after a write, it raises an OSError.

It's not yet clear to me why this happens. I wrote some some small tests in pyfilesystem2 that reproduce it and I'll try to get back to you with a better explanation. I just figured you might want to know.

sqwishy:
I'm super dumb. The file should be opened with `os.O_WRONLY` or one of those flags. I will open a new pull request for this.

PR #190, opened 2018-07-17
Fix parameter docs for `FS.settext`

Hoboneer:
Match with the docs for `FS.setbytes` because they're different versions
of essentially the same thing: setX to a file, so `path` is exactly the same
and `contents` is adapted for this function.


willmcgugan:
Thanks.

Issue #192, opened 2018-07-20
Case insensitive walk


geoffjukes:
Hi!,

I'm walking a case insensitive filesystem, with a filter: `fs.open_fs(SCAN).walk.files(filter=['*.wav']`

It doesn't return files ending `.WAV` because the filter is case-sensitive.

Is there an option to make that case insensitive? I'm happy to check the code and put in a PR if it doesn't already exist (I couldn't find anything in the docs)

willmcgugan:
That *should* work. `walk` will call `fs.match` which will detect if the filesystem is case sensitive with the `case_sensitive` meta value. Although it may not be a very well tested mechanism.

What filesystem is it you are using?

geoffjukes:
I’m using a case insensitive SMB share, mounted in a Linux host.

willmcgugan:
I've just noticed a bug. The `fs.match` method references a meta value called `case_sensitive`, but it is `case_insensitive` that is used everywhere.

Feel free to tackle that one. Alternatively, you could do `filter=['*.wav', '*.WAV']` until I fix it...

geoffjukes:
Good spot - I'm on it :)

geoffjukes:
#193

willmcgugan:
Should be fixed in 2.0.26.

PR #193, opened 2018-07-20
Fix `case_sensitive` bug #192

geoffjukes:
Fixes bug in `get_matcher` logic that referenced `case_sensitive` rather than `case_insensitive`.

To maintain consistency, I refactored all references of `case_sensitive` to `case_insensitive`, inverting defaults, inputs and docstrings as appropriate.

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -1552,9 +1552,9 @@ class FS(object):
             return True
         if isinstance(patterns, six.text_type):
             raise TypeError('patterns must be a list or sequence')
-        case_sensitive = typing.cast(
-            bool, self.getmeta().get('case_sensitive', True))
-        matcher = wildcard.get_matcher(patterns, case_sensitive)
+        case_insensitive = typing.cast(
+            bool, self.getmeta().get('case_insensitive', False))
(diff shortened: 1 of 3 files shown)


willmcgugan:
I see you were trying to make the naming consistent here, but this has changed the public api -- which could conceivably break other projects.

I think we should stick with `case_sensitive` in the method signatures, because it reads like a double negative when you have to do `if not case_insensitive`, and I don't like no double negatives.

I think I called the meta value `case_insensitive` because I like absense of a meta key to be mean false, and case sensitive feels like the default. Might have to live with that inconsistency, but at least its only relevant in one method.

If you don't mind, I'm going to tackle this one myself since I have a little time atm.

geoffjukes:
“don’t like no double negatives” 🤣

No worries at all. I did consider the change in API, and check a few of the other filesystems out there, and found no conflict (all seemed to rely on the meta value) but the concern is reasonable. 

The other option I considered was just to fix the typo and invert the result - but that just seemed waaaay too ugly 😀 

```
not self.getmeta().get('case_insensitive', False))
```

PR #194, opened 2018-07-22
Typing

willmcgugan:
Some fixes for typing with latests MyPy.

### DIFF
--- a/fs/_fscompat.py
+++ b/fs/_fscompat.py
@@ -5,7 +5,7 @@ import six
 try:
     from os import fsencode, fsdecode
 except ImportError:
-    from backports.os import fsencode, fsdecode
+    from backports.os import fsencode, fsdecode  # type: ignore
 
 try:
     from os import fspath
--- a/fs/_typing.py
+++ b/fs/_typing.py
@@ -8,12 +8,11 @@ import six
 
 _PY = sys.version_info
 
+from typing import overload  # type: ignore
 
 if _PY.major == 3 and _PY.minor == 5 and _PY.micro in (0, 1):
     def overload(func):  # pragma: no cover
         return func
-else:
-    from typing import overload  # type: ignore
 
 try:
     from typing import Text
--- a/fs/_typing.py
+++ b/fs/_typing.py
@@ -13,9 +13,9 @@ if _PY.major == 3 and _PY.minor == 5 and _PY.micro in (0, 1):
     def overload(func):  # pragma: no cover
         return func
 else:
-    from typing import overload
+    from typing import overload  # type: ignore
 
 try:
     from typing import Text
 except ImportError:  # pragma: no cover
-    Text = six.text_type
+    Text = six.text_type  # type: ignore
--- a/fs/compress.py
+++ b/fs/compress.py
@@ -90,7 +90,8 @@ def write_zip(src_fs,                            # type: FS
             zip_info = zipfile.ZipInfo(zip_name, zip_time)  # type: ignore
 
             try:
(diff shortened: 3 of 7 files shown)


fs/_fscompat.py
```diff
@@ -5,7 +5,7 @@
 try:
     from os import fsencode, fsdecode
 except ImportError:
-    from backports.os import fsencode, fsdecode
+    from backports.os import fsencode, fsdecode  # type: ignore
```
willmcgugan: Mypy doesn't seem to complain about this.

fs/_typing.py
```diff
@@ -8,14 +8,13 @@
 
 _PY = sys.version_info
 
+from typing import overload  # type: ignore
```
willmcgugan: This fixed some baffling errors with overload.

fs/compress.py
```diff
@@ -90,7 +90,8 @@ def write_zip(src_fs,                            # type: FS
             zip_info = zipfile.ZipInfo(zip_name, zip_time)  # type: ignore
 
             try:
-                zip_info.external_attr = info.permissions.mode << 16
+                if info.permissions:
```
willmcgugan: `info.permissions` could be None according to the typing information. I'm not certain if it could be None in practice.

Perhaps a cast would be more appropriate there?

fs/ftpfs.py
```diff
@@ -82,7 +83,7 @@ def ftp_errors(fs, path=None):
                 msg=message
             )
         elif code in ('501', '550'):
-            raise errors.ResourceNotFound(path=path)
+            raise errors.ResourceNotFound(path=cast(str, path))
```
willmcgugan: `path` may be None, but in practice this error should only occur if there is a path.

fs/info.py
```diff
@@ -82,23 +82,18 @@ def _make_datetime(self, t):
...
 
-    @overload
-    def get(self, namespace, key, default=None):  # pragma: no cover
-        # type: (Text, Text, Optional[T]) -> Optional[T]
-        pass
-
     @overload
     def get(self, namespace, key):  # pragma: no cover
-        # type: (Text, Text) -> Optional[T]
+        # type: (Text, Text) -> Any
```
willmcgugan: Mypy wasn't having this. The return type was more general than the original code implied.

It would actually me more accurate to replace Any with a union of str, int, list, dict etc, but that would require a cast everywhere `get` is used. Could still be persuaded to do that...

fs/move.py
```diff
@@ -16,7 +16,7 @@
 
 
 def move_fs(src_fs, dst_fs, workers=0):
-    # type: (Union[Text, FS], Union[Text, FS]) -> None
+    # type: (Union[Text, FS], Union[Text, FS], int) -> None
```
willmcgugan: forgot the workers type

fs/compress.py
althonos: I'm not certain, but what about a filesystem with a concept of ownership but no notion of permissions ? It would then only define the `access` namespace partially. IMO we should not make the assumption that `info.permissions` is never `None`.

(Shouldn't it you use `if info.permissions is not None` just to be explicit ? I had to look to the `fs.permissions` source to check that `Permissions.__bool__` was not implemented)

fs/ftpfs.py
althonos: Good catch !

fs/compress.py
willmcgugan: Good point. Should definitely be an explicit check for None.

PR #198, opened 2018-07-24
Fix tests against Python 3.7

althonos:
This time I tested using the Docker `python:3.7-alpine` image, so hopefully this should be fine.

```shell
Ran 1513 tests in 42.330s

OK (passes=1488, skips=25)
```


willmcgugan:
Thanks

Issue #205, opened 2018-07-31
Error walking ftp filesystem


dhirschfeld:
When trying to walk an FTP sub-folder as shown below:

![image](https://user-images.githubusercontent.com/881019/43458119-5a4f05c0-950c-11e8-8a3c-da9a57d570ed.png)

The below code:
https://github.com/PyFilesystem/pyfilesystem2/blob/b2c0d96f55d4dfe777b4f9476676b77d01f36bf7/fs/ftpfs.py#L710-L724

...seems to give sensible results:
```python
In[2]: lines
Out[2]: 
['Type=cdir;Modify=20180731114724;UNIX.mode=0755; /tmp',
 'Type=pdir;Modify=20180731112024;UNIX.mode=0775; /',
 'Type=file;Size=331523;Modify=20180731112041;UNIX.mode=0644; a.csv',
 'Type=file;Size=368340;Modify=20180731112041;UNIX.mode=0644; b.csv']

In[4]: list(self._parse_mlsx(lines))
Out[4]: 
[{'basic': {'name': 'tmp', 'is_dir': True},
  'ftp': {'type': 'cdir', 'modify': '20180731114724', 'unix.mode': '0755'},
  'details': {'type': 1, 'size': 0, 'modified': 1533037644}},
 {'basic': {'name': '', 'is_dir': True},
  'ftp': {'type': 'pdir', 'modify': '20180731112024', 'unix.mode': '0775'},
  'details': {'type': 1, 'size': 0, 'modified': 1533036024}},
 {'basic': {'name': 'a.csv', 'is_dir': False},
  'ftp': {'type': 'file',
   'size': '331523',
   'modify': '20180731112041',
   'unix.mode': '0644'},
  'details': {'type': 2, 'size': 331523, 'modified': 1533036041}},
 {'basic': {'name': 'b.csv', 'is_dir': False},
  'ftp': {'type': 'file',
   'size': '368340',
   'modify': '20180731112041',
   'unix.mode': '0644'},
  'details': {'type': 2, 'size': 368340, 'modified': 1533036041}}]
```
i.e. the current dir, the parent dir and the two files are all correctly? listed.

However, in the [`Walker._walk_breadth`](https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/walk.py#L389-L415) method the current dir `tmp` is the first item yielded from `self._scan` which result in the `cdir` being joined to itself and the non-existent `/tmp/tmp` being pushed onto the deque:

[...]

PR #208, opened 2018-08-05
fix for ftp dir listing

willmcgugan:
Fix for https://github.com/PyFilesystem/pyfilesystem2/issues/205


dhirschfeld:
It was on my TODO to fixup my initial PR but I'm totally swamped so thanks for the fix! 👍

PR #209, opened 2018-08-05
implementation of glob

willmcgugan:
# WIP

An implementation of 'glob' for Pyfilesystem.

This finds all Python files in the root directory:

```python
for path, info my_fs.glob("*.py"):
    print(path)
```

This finds all Python files in sub directories:

```python
for path, info my_fs.glob("*/*.py"):
    print(path)
```

This finds all Python files recursively:

```python
for path, info my_fs.glob("**/*.py"):
    print(path)
```

This finds all the .git directories:

```python
for path, info in my_fs.glob("**/.git/"):
    print(path)
```

Similar to `walk`, `glob` is a callable object, which you can either iterate over or call its methods.

This counts the number of bytes in Python files recursively:

```python
print(my_fs.glob("**/*.py").count())
```

This removes all `pyc` file recursively:

```python
my_fs.glob("**/*.pyc").remove()
```

Many other combinations are possible, I'm sure.

All of this can be done already, but its a nice syntactical sugar IMHO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.1.0] - Unreleased
+
+### Added
+
+- fs.glob support
+
 ## [2.0.27] - 2018-08-05
 
 ### Fixed
@@ -159,6 +165,7 @@ No changes, pushed wrong branch to PyPi.
 ## [2.0.8] - 2017-08-13
 
 ### Added
+
 - Lstat info namespace
 - Link info namespace
 - FS.islink method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [2.1.0] - Unreleased
+## [2.1.0] - 2018-08-12
 
 ### Added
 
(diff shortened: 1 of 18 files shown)


willmcgugan:
@geoffjukes @althonos Any thoughts on this one? Interface is still very much up for debate.

willmcgugan:
This makes the example on https://www.pyfilesystem.org shorter...

```python
def count_python_loc(fs):
    """Count non-blank lines of Python code."""
    count = 0
    for path in fs.walk.files(filter=['*.py']):
        with fs.open(path) as python_file:
            count += sum(1 for line in python_file if line.strip())
    return count
```

```python
def count_python_loc(fs):
    """Count non-blank lines of Python code."""
    return fs.glob("**/*.py").count_lines().non_blank
```

geoffjukes:
@willmcgugan this looks really nice. I hadn't come across the 'double asterisk' glob before, strangely.

I think this a useful addition, especially for operations that would previously have required a single-item filter (of which I have several)

althonos:
The glob asterisk I tend to use a lot in `fish`, so that looks really good to me. Maybe we could support all the [UNIX globbing patterns](https://en.wikipedia.org/wiki/Glob_(programming)), , so that UNIX users would feel right at home even on Windows ? :smile:

geoffjukes:
@willmcgugan Looking forward to this becoming a release :)

willmcgugan:
@geoffjukes It's already in!

willmcgugan:
Ah. The lack of real folders in S3FS is a headache for me. I couldn't see a way of making it appear like there are real folders that isn't a performance killer.

Btw, what is it you tried exactly? I proposed adding `remove` and friends to walker objects, but settled on a glob object which does the same thing. i.e.

```python
fs.glob('**/*.pyc').remove()
```

geoffjukes:
I had tried the remove-on-walker. I honestly didn’t read the docs, I just read the original ticket and rolled with it. I’ll check the glob method tomorrow, it light actually work for me.

I’ll also retry what I need using the glob on the bucket. IMHO, then deleting of a ‘folder’ in S3 should just silently skip if the ‘/‘ file isn’t there. For my needs, I just set up a Boto3 pagination and list the keys with a prefix. I only allow a ‘one or all’ delete option (no matching). It satisfies my requirement.


________________________________

PR #211, opened 2018-08-21
minor improvements and cleanup at fs.test.test_listdir

birnbaum:
Few minor improvement of `fs.test.test_listdir`:

1. Listing over files _and_ a directory. This would have caught a bug I int in my GCSFS implementation: A `fs.listdir`on the new example would have resulted in `["foo", "bar", "baz/"]` which is incorrect.
2. Added a file in a subdirectory which should not be listed (easy mistake when trying to implement an object store where every "list" is actually a "walk"). This was caught by other unit tests but I think the check makes sense here.
3. The unicode check was duplicated, the same check is done a few lines further down

### DIFF
--- a/fs/test.py
+++ b/fs/test.py
@@ -543,14 +543,13 @@ class FSTestCases(object):
         self.assertEqual(self.fs.listdir("."), [])
         self.assertEqual(self.fs.listdir("./"), [])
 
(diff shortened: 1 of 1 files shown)


willmcgugan:
Thanks.

PR #215, opened 2018-09-05
Use os.O_RDWR for exclusive mode in Python 2.7

sqwishy:
Also doing the exclusive thing properly in openbin.

(#182)

I think the OS_RDWR behaviour is correct since if you specify `x` mode, you can't specify `r` or `w`. So it implies both. I'm pretty sure this is what Python 3 does so the `fs` package API should be consistent in both versions of Python here.


willmcgugan:
Thanks!

PR #216, opened 2018-09-17
Implement OSFS.copy using sendfile or shutil.copy

althonos:
Hi Will,

while toying with the `sendfile` syscall I noticed that it could also be used to copy files on the OS filesystem quicker than with a chunk copy, and of course people already patched the `shutil` module to use it (see [python#33671](https://bugs.python.org/issue33671)). 

I backported the patch of [`pyfastcopy`](https://pypi.org/project/pyfastcopy/) so that `OSFS.copy` will use `sendfile` if it is available (as either `os.sendfile` or `sendfile.sendfile` from [`pysendfile`](https://pypi.org/project/pysendfile)) and falling back to `shutil.copyfile` if it is not supported on the local platform.

It may also be a good idea to use `sendfile` and `shutil.copy` in `fs.copy.copy_file` and `fs.copy.copy_file_internal` if both source and destination files have a syspath. Tell me what you think so I could get that done before merging !

(I'm also working on getting `FTPFS.setbinfile` use `sendfile` but I'll add a separate PR for that because it needs more boilerplate to ensure compatibility on all platforms and Python versions).

### DIFF
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -29,11 +29,13 @@ except ImportError:
     except ImportError:  # pragma: no cover
         scandir = None
 
-# NB: backport of https://bugs.python.org/issue33671 patch
 try:
-    import pyfastcopy
-except ImportError:  # pragma: no cover
-    pass
+    from os import sendfile
+except ImportError:
+    try:
+        from sendfile import sendfile
+    except ImportError:
+        sendfile = None
 
 from . import errors
 from .errors import FileExists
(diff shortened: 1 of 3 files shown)


althonos:
@willmcgugan : since `os.sendfile` is already in the standard library from Python 3.5 onwards on UNIX platforms, the `TestOSFS.test_copy` will be testing the `sendfile` code if available. Given the current build matrix it means all the possible cases are tested (check coveralls; the only line I missed checks for the existence of the parent directory of the destination file, I'll add a test for that).

I also added a test that checks that the fallback implementation is used when `sendfile` is present but fails.

fs/osfs.py
```diff
@@ -361,90 +370,85 @@ def removedir(self, path):
...
+            errno.EOPNOTSUPP,
+        })
+
+        @staticmethod
+        def _copy_shutil(src_path, dst_path):
+            return shutil.copy2(src_path, dst_path)
+
+        @classmethod
+        def _copy_sendfile(cls, src_path, dst_path):
+            sent = maxsize = 2**31 - 1
```
willmcgugan: Any particular reason to copy 2GB at a time?

fs/osfs.py
```diff
@@ -361,90 +370,85 @@ def removedir(self, path):
...
+            offset = 0
+            with io.open(src_path) as src:
+                with io.open(dst_path, 'w') as dst:
+                    fd_src, fd_dst = src.fileno(), dst.fileno()
+                    try:
+                        while sent > 0:
```
willmcgugan: Have you followed the assignment expressions pep? 

One day you could write this:

```python
while sent := sendfile(fs_dst, fd_src, offset, maxsize):
    offset += sent
```

althonos: I know, I hate the syntax but I love the usecase ! ^_^
(I really can't figure out why they didn't use the `as` keyword instead of introducing this `:=` monstrosity :smile: )

fs/osfs.py
althonos: I think in the implementation I got that from they used that to ensure `sendfile` would be trying to send as much data as possible. It may be a better idea to replace that with an `os.fstat` call to get the filesize.

willmcgugan:
Test pass on OSX. Thanks.

PR #226, opened 2018-10-24
Issue 225

televi:
Tested by:
* running tox on my system on master and then comparing that with my branch (I saw the zipfs tests fail in both).  No new failures.
* runnining python3.6 setup.py sdist - build was successful
* unzip the above sdist - LICENSE file was present
* run pip wheel bdist_wheel from the above unzipped sdist - wheel built fine

I didn't include any new tests since this is just build time changes and nothing operational - let me know if that is not OK.


televi:
Not sure what to do about the py3.3 failure - it "claims" pip==7.1.2 was installed but the error is not from that version of pip - that file doesn't even exist in pip 7.1.2 (but it does exist in later versions of pip).

televi:
Looking a bit deeper... what it really installed was pip==18.1.  In the source for that, they stopped supporting py3.3 (that's the error at line 90 in the output).  In pip==18.0, it still has support for py3.3, so it looks like the travis config will need to pin pip <= 18.0 for py3.3

willmcgugan:
Thanks @televi. @althonos do you want to give this a quick scan?

willmcgugan:
Thanks @televi

althonos:
@willmcgugan : sorry I didn't see you had mentioned me !

PR #232, opened 2018-11-14
Revert back to manual registry installation (fix #228)

### LINKED ISSUE #228: Evaluated options for opener discovery
I'm not sure this is indeed an `fs` issue, but perhaps it's at least good to have it documented here, too.

We're running into a problem with embedding `fs` into an application for MacOS with py2app, where the opener registry can't find anything, because the `entry_points.txt` file is not copied into the application, so `pkg_resources.iter_entry_points("fs.opener")` yields nothing.

See the related issue for py2app:
https://bitbucket.org/ronaldoussoren/py2app/issues/256/fs-module-not-fully-working-from-app

It seems the use of package metadata like `entry_points.txt` is problematic, and I wonder if there isn't a more friendly way to populate a registry such as `fs.opener`.

Cc @typemytype

althonos:
Hi @willmcgugan, hi @justvanrossum,

this PR reverts back to the original Pyfilesystem2 opener behaviour, and manually installs openers from the `fs.opener` module. Additional extensions are loaded, and take priority over builtin filesystems (i.e. if an extension defines `tar` then it is used instead of the `fs.opener.tarfs.TarOpener`).

I added an optional kwarg to `Registry`, so that a `Registry` can disable loading external filesystems.  But because of variable shadowing, no `Registry` instance can be created outside of the `fs.opener.registry` module anyway (well, it's still possible through `sys.modules`).

Tests still passed on my machine with `fs.sshfs` installed.

### DIFF
--- a/fs/opener/__init__.py
+++ b/fs/opener/__init__.py
@@ -10,6 +10,9 @@ from .base import Opener
 from .parse import parse_fs_url as parse
 from .registry import registry
 
+# Import opener modules so that `registry.install` if called on each opener
+from . import appfs, ftpfs, memoryfs, osfs, tarfs, tempfs, zipfs
+
 # Alias functions defined as Registry methods
 open_fs = registry.open_fs
 open = registry.open
--- a/fs/opener/appfs.py
+++ b/fs/opener/appfs.py
@@ -35,6 +35,7 @@ class AppFSOpener(Opener):
         create,  # type: bool
         cwd,  # type: Text
     ):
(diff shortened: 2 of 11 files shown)


justvanrossum:
Wow fantastic! I'll see if I can test this locally soon. Cc @typemytype.

willmcgugan:
Thanks @althonos Give me a few days to look this over.

justvanrossum:
Sorry for my slow reply. I just tested this and it works just fine! +1 from me to merge this. Thanks so much for working on this.

willmcgugan:
Thanks @althonos

PR #234, opened 2018-11-21
Fix `AppFS` failing to create missing dirs even on `create=True`

althonos:
Currently, `AppFS` won't create a new subdirectory even if `create` is set to `True`. Minimal failing example:

```python
import fs
fs.open_fs("userdata://fs:willmcgugan/test", create=True)
```

As a side note: maybe the `fs.base.FS.makedir` and `fs.base.FS.makedirs` could also take a `factory` argument like `fs.base.FS.opendir` does !

### DIFF
--- a/fs/opener/appfs.py
+++ b/fs/opener/appfs.py
@@ -61,6 +61,8 @@ class AppFSOpener(Opener):
         app_fs = fs_class(appname, author=author, version=version, create=create)
 
         if delim:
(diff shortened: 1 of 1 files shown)


willmcgugan:
Thanks!

I see what you mean about adding `factory` to makedir(s). It would be cleaner just to return makedir result. Probably worth doing at some point.

PR #248, opened 2018-12-29
experimental api refresh

willmcgugan:
# API Refresh (WIP)

I've had at the back of my mind for a while that some of the api methods could be better named. I think it was @althonos that first pointed out some inconsistencies in the naming. This PR is an experimental branch where I can test potential method name changes. I'll elicit feedback before I merge any of this.

~It will be a breaking change, so it would require a major version increment, according to semver. Hopefully everyone has learned to pin their requirements!~

Added the old methods as aliases with deprecation warnings. So it isn't a breaking change yet.

The changes are mostly in the helper methods, which most filesystems don't implement. I wouldn't expect many third party filesystems to have to make changes other than updating requirements. Applications _could_ be affected though.

Please see comments in base.py


fs/base.py
```diff
@@ -550,7 +550,7 @@ def exclude_file(patterns, info):
             iter_info = itertools.islice(iter_info, start, end)
         return iter_info
 
-    def getbytes(self, path):
+    def readbytes(self, path):
```
willmcgugan: get/set would be better as read/write for filesystem parlance.

fs/base.py
```diff
@@ -568,7 +568,7 @@ def getbytes(self, path):
             contents = read_file.read()
         return contents
 
-    def getfile(self, path, file, chunk_size=None, **options):
+    def download(self, path, file, chunk_size=None, **options):
```
willmcgugan: `getfile` has always felt awkward as it doesn't *return* a file, like `gettext` returns text.

From a network filesystem this makes total sense, it may not be quite so obvious for other filesystems, but I think it actually fits.

fs/base.py
```diff
@@ -1244,7 +1244,7 @@ def setbytes(self, path, contents):
         with closing(self.open(path, mode="wb")) as write_file:
             write_file.write(contents)
 
-    def setbinfile(self, path, file):
+    def upload(self, path, file):
```
willmcgugan: Like `download`, I think `upload` better describes what this does than the clumsy `setbinfile`.

willmcgugan:
@althonos @zopyx @geoffjukes @birnbaum @lurch Looking for feedback on this one. Not in a hurry to merge. Please see description above.

zopyx:
I dont care much about the name changes. Perhaps you should provide aliases and deprecation warnings for older code.

willmcgugan:
@zopyx Good idea re deprecation warnings. Then it wouldn't require a major version change straight away.

althonos:
I'm happy with all the changes: the more concerning issue to me was the inconsistent names of `setfile`, `setbinfile` and `getfile`. Having `upload`/`download` feels a lot better.

fs/ftpfs.py
```diff
@@ -703,9 +701,7 @@ def removedir(self, path):
                 raise  # pragma: no cover
 
     def _scandir(
-        self,
-        path,  # type: Text
-        namespaces=None   # type: Optional[Container[Text]]
+        self, path, namespaces=None  # type: Text  # type: Optional[Container[Text]]
```
willmcgugan: Black broke mymy comments. Todo

geoffjukes:
Thanks @willmcgugan and @althonos for updating the other packages (s3 & Archive) for this too. It made this a simple update for me (only a handful of references). 👍

PR #249, opened 2019-01-05
Fix `registry.install` not returning decorated class

althonos:
# PyFilesystem Pull Request

Thank you for your pull request!

## Type of changes

- [x] Bug fix
- [ ] New Feature
- [ ] Other

## Checklist

- [x] I've run the latest [black](https://github.com/ambv/black) with default args on new code
- [x] I've updated CHANGELOG.md
- [x] I accept that @willmcgugan may be pedantic in the code review

## Description

I made sure `Registry.install` returns its argument, otherwise it would not be possible to use it as a decorator as demonstrated in the docstring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- `Registry.install` returns its argument.
+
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks. I'll add a test for this.

willmcgugan:
Actually, that wouldn't have broken anything from the registry point of view. The decorator still manages to store a reference to the class. It would have made the opener unimportable as it would have replaced the class with `None` in the module namespace. Which I guess is why it went undetected!

willmcgugan:
btw I release 2.2.1 with that fixed.

Issue #262, opened 2019-02-10
non-ascii characters not accepted in filenames


simonm3:
The following copy_file fails. Same with copy_dir if a filename contains non-ascii characters.

`copy_file(s3, "job-data/job-453/13 – Marked Register.pdf", local, "job-data/job-453/13 – Marked Register.pdf")
`
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-13-c5d4e6cd8643> in <module>
----> 1 copy_file(s3, "job-data/job-453/13 – Marked Register.pdf", local, "job-data/job-453/13 – Marked Register.pdf")

/anaconda3/lib/python3.6/site-packages/fs/copy.py in copy_file(src_fs, src_path, dst_fs, dst_path)
    138                 with _src_fs.lock(), _dst_fs.lock():
    139                     if _dst_fs.hassyspath(dst_path):
--> 140                         with _dst_fs.openbin(dst_path, "w") as write_file:
    141                             _src_fs.getfile(src_path, write_file)
    142                     else:

/anaconda3/lib/python3.6/site-packages/fs/osfs.py in openbin(self, path, mode, buffering, **options)
    320         self.check()
    321         _path = self.validatepath(path)
--> 322         sys_path = self._to_sys_path(_path)
    323         with convert_os_errors("openbin", path):
    324             if six.PY2 and _mode.exclusive and self.exists(path):

/anaconda3/lib/python3.6/site-packages/fs/osfs.py in _to_sys_path(self, path)
    164         """
    165         sys_path = fsencode(
--> 166             os.path.join(self._root_path, path.lstrip("/").replace("/", os.sep))
    167         )
    168         return sys_path

/anaconda3/lib/python3.6/os.py in fsencode(filename)
    800         filename = fspath(filename)  # Does type-checking of `filename`.
    801         if isinstance(filename, str):
--> 802             return filename.encode(encoding, errors)
    803         else:
    804             return filename

UnicodeEncodeError: 'ascii' codec can't encode character '\u2013' in position 24: ordinal not in range(128)

willmcgugan:
Hi,

For some reason it's failing to encoding the mdash character in the filename. What is strange is that it is encoding it as ascii. 

What platform is this on?

Could you try the following and let me know what the result is:

```python
>>> import sys
>>> sys.getfilesystemencoding()
```

Could you also let me know what is in your environment:

```python
>>> import os
>>> os.environ
```

simonm3:
[...]

PR #263, opened 2019-02-10
check invalid encoding in OSFS

willmcgugan:
# WIP

## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

When there are encoding issues in OSFS, a unicode decode error is raised. This can often be cryptic and breaks the promise that FS objects raise FS errors.

I've customized `OSFS.validatepath` to attempt a `fsencode` and raise a `InvalidCharsInPath` error when encoding fails. The error message also suggest to look at the env var LANG, which is a likely cause of encoding fails.


willmcgugan:
Prompted by https://github.com/PyFilesystem/pyfilesystem2/issues/262

PR #265, opened 2019-02-15
Walk filter

willmcgugan:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Added an `exclude` argument to the walker, which excludes files matching any of the given patterns. This can be useful to ignore temporary or hidden files.

Also added `filter_dirs` argument to the walker, which will only recurse those directories which match any of the given filters. This one was mainly added for parity with other filter / exclude arguments. I can't actually think of a use case for this functionality!

### DIFF
--- a/fs/walk.py
+++ b/fs/walk.py
@@ -539,9 +539,12 @@ class BoundWalker(typing.Generic[_F]):
             search (str): If ``'breadth'`` then the directory will be
                 walked *top down*. Set to ``'depth'`` to walk *bottom up*.
             filter (list): If supplied, this parameter should be a list
-                of file name patterns, e.g. ``['~*', '.*']``. Files will only be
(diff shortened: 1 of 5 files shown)


fs/walk.py
```diff
@@ -398,19 +416,27 @@ def _walk_breadth(
         queue = deque([path])
         push = queue.appendleft
         pop = queue.pop
-        depth = self._calculate_depth(path)
+
+        _combine = combine
```
willmcgugan: Micro-optimizations. Locals are faster than an attribute on self, which requires a dict lookup.

TBH the difference may only be noticeable on benchmarks. In the real world where work is done on the fs being walked, these speed-ups will be negligible.

Issue #256, opened 2019-01-22
ZipFS (Writing Mode) Error on Windows


merlink01:
Python 35 (32Bit)
Pyfilesystem: 2.2.1

Testcode:
```
from fs import zipfs
testfs = zipfs.ZipFS('test.zip',write=True)
testfile = testfs.open('testfile.txt','w')
testfile.write('Hello World')
testfile.close()
testfs.close()
```

Result: An empty ZipFile is created.

After looking in the code I saw a Try,Finally around 
self.write_zip() (zipfs.py:239)

This is really bad, because here a Silent Error + Data Loss happened.

After disabling the Try Finally I got this Error:

```
Traceback (most recent call last):
  File "error.py", line 7, in <module>
    testfs.close()
  File "C:\Python35-32\lib\site-packages\fs\zipfs.py", line 239, in close
    self.write_zip()
  File "C:\Python35-32\lib\site-packages\fs\zipfs.py", line 271, in write_zip
    encoding=encoding or self.encoding,
  File "C:\Python35-32\lib\site-packages\fs\compress.py", line 61, in write_zip
    for path, info in gen_walk:
  File "C:\Python35-32\lib\site-packages\fs\walk.py", line 385, in info
    for _path, info in _walk:
  File "C:\Python35-32\lib\site-packages\fs\walk.py", line 405, in _walk_breadth
    for info in self._scan(fs, dir_path, namespaces=namespaces):
  File "C:\Python35-32\lib\site-packages\fs\walk.py", line 277, in _scan
    for info in fs.scandir(dir_path, namespaces=namespaces):
  File "C:\Python35-32\lib\site-packages\fs\osfs.py", line 459, in _scandir
    for dir_entry in scandir(sys_path):
TypeError: os.scandir() doesn't support bytes path on Windows, use Unicode instead
```

Any Idea what I can do to get a zipfile created?

merlink01:
OK, Tested on Python 37 (32Bit) and it works.

willmcgugan:
Not sure why the try/finally would hide any errors there.

I think the issue boils down to this section I found in the docs:

> On Unix, path can be of type str or bytes (use fsencode() and fsdecode() to encode and decode bytes paths). On Windows, path must be of type str. On both systems, the type of the name and path attributes of each DirEntry will be of the same type as path.

So it looks like Windows may need a special case there.

I've tagged this as a bug. In the short term, you could try a 'pip uninstall scandir'.

[...]

PR #269, opened 2019-02-22
Win fix

willmcgugan:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [X] Tests
- [ ] Other

## Checklist

## Description

Fixes https://github.com/PyFilesystem/pyfilesystem2/issues/256 and a number of other Windows issues. Now that we have Windows set up in appveyor, we can keep on top of Windows support

### DIFF
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -12,44 +12,48 @@ import fs
 from fs.osfs import OSFS
 
 
-@unittest.skipIf(platform.system() == "Darwin", "Bad unicode not possible on OSX")
-@unittest.skipIf(platform.system() == "Windows", "Strict encoding on Windows")
-class TestEncoding(unittest.TestCase):
+if platform.system() != "Windows":
 
-    TEST_FILENAME = b"foo\xb1bar"
-    TEST_FILENAME_UNICODE = fs.fsdecode(TEST_FILENAME)
+    @unittest.skipIf(platform.system() == "Darwin", "Bad unicode not possible on OSX")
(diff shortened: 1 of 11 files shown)


tests/test_encoding.py
```diff
@@ -12,43 +12,48 @@
...
-            dirlist = test_fs.listdir("/")
-            self.assertEqual(dirlist, [self.TEST_FILENAME_UNICODE])
-            self.assertEqual(test_fs.readtext(dirlist[0]), "baz")
-
-    def test_scandir(self):
-        with OSFS(self.dir_path) as test_fs:
-            for info in test_fs.scandir("/"):
-                self.assertIsInstance(info.name, six.text_type)
-                self.assertEqual(info.name, self.TEST_FILENAME_UNICODE)
+if platform.system() != "Windows":
```
althonos: Why not wrap in two `unittest.skipIf` decorators with two different conditions ?

willmcgugan: It was breaking at the module level. Fsencode uses 'strict' error handler and was throwing an exception on the classvar.

althonos: @willmcgugan : my bad !

willmcgugan:
No worries! I've got Windows tests running with appveyor now, so we can catch these things in future.

tests/test_encoding.py
willmcgugan: No worries! I've got Windows tests running with appveyor now, so we can catch these things in future.

Issue #284, opened 2019-05-10
another exception when a tarfile contains a path called '.' 


mrg0029:
I was running into [this](https://github.com/PyFilesystem/pyfilesystem2/issues/275) problem until the 2.4.5 release.
I no longer get an IndexError, but I now get a ResourceNotFound when testing the same file.
In Debian/Ubuntu/etc. this should reproduce it:

    $ pushd $(mktemp -d)
    $ apt-get download hello
    $ ar t hello*
        debian-binary
        control.tar.gz
        data.tar.xz
    $ ar x hello*
    $ python -c 'import tarfile; tarfile.open("data.tar.xz").getmembers(); print("OK")'
        OK

    $ python -c 'import fs; fs.open_fs("tar://data.tar.xz").walk.files().__next__(); print("OK")'
        Traceback (most recent call last):
          File "site-packages/fs/tarfs.py", line 330, in getinfo
            member = self._tar.getmember(self._encode(_path))
          File "tarfile.py", line 1752, in getmember
            raise KeyError("filename %r not found" % name)
        KeyError: "filename 'usr' not found"

        During handling of the above exception, another exception occurred:

        Traceback (most recent call last):
          File "<string>", line 1, in <module>
          File "site-packages/fs/walk.py", line 362, in files
            for _path, info in self._iter_walk(fs, path=path):
          File "site-packages/fs/walk.py", line 433, in _walk_breadth
            for info in _scan(fs, dir_path, namespaces=namespaces):
          File "site-packages/fs/walk.py", line 298, in _scan
            six.reraise(type(error), error)
          File "site-packages/six.py", line 692, in reraise
            raise value.with_traceback(tb)
        fs.errors.ResourceNotFound: resource '/usr' not found

I don't get the same error when creating a file manually.

    $ pushd $(mktemp -d)
    $ touch a b c
    $ tar --create      -f stuff1.tar    a b c
    $ tar --create --xz -f stuff2.tar.xz a b c
    $ python -c 'import fs; fs.open_fs("tar://stuff1.tar").walk.files().__next__(); print("OK")'
        OK
    $ python -c 'import fs; fs.open_fs("tar://stuff2.tar.xz").walk.files().__next__(); print("OK")'
        OK

Edit: fixed a couple of the commands per lurch's reply

lurch:
`tar tvf data.tar.gz` shows:
```
drwxr-xr-x root/root         0 2013-08-26 17:25 ./
[...]

PR #291, opened 2019-06-02
Tar pit

willmcgugan:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [x] Other

## Checklist

## Description

Fixes https://github.com/PyFilesystem/pyfilesystem2/issues/284

Also dropped Python3.3 support to solve Github security warning. We support Python3.4 - Python3.7 and 3.8 is around the corner, so I think its time. Hope there is no pushback to that.

### DIFF
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -338,7 +338,7 @@ class ReadTarFS(FS):
         else:
             try:
                 implicit = False
-                member = self._tar.getmember(self._encode(_path))
+                member = self._directory[_path]
             except KeyError:
                 if not self.isdir(_path):
                     raise errors.ResourceNotFound(path)
@@ -426,7 +426,7 @@ class ReadTarFS(FS):
             raise errors.ResourceReadOnly(path)
 
         try:
-            member = self._tar.getmember(self._encode(_path))
+            member = self._directory[_path]
         except KeyError:
             six.raise_from(errors.ResourceNotFound(path), None)
 
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -276,14 +276,16 @@ class ReadTarFS(FS):
         self._directory_cache = None
 
     @property
-    def _directory(self):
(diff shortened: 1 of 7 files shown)


fs/tarfs.py
```diff
@@ -280,9 +281,19 @@ def _directory(self):
         if self._directory_cache is None:
             _decode = self._decode
             _directory = ((_decode(info.name).strip("/"), info) for info in self._tar)
```
lurch: When I was debugging the linked issue, I found the `_directory` var-name very confusing. Perhaps `_directory_entries` or something would be better?

fs/tarfs.py
```diff
@@ -280,9 +281,19 @@ def _directory(self):
...
-                (name, info) for name, info in _directory if normpath(name)
-            )
+
+            def _list_tar():
+                for name, info in _directory:
+                    try:
+                        _name = normpath(name)
+                    except IllegalBackReference:
+                        # Back references outside root, must be up to no good.
+                        pass
```
lurch: Rather than _always_ silently ignoring back-reference errors, I wonder if it'd be worth ignoring them by default, with an FS init-option to re-throw the back-reference errors?

fs/tarfs.py
willmcgugan: Good point

fs/tarfs.py
willmcgugan: I pondered that, but if we raise an error at this point then a tar with a single badly formatted entry would prevent the tar from being opened. At least this way you could access any other valid entries via the TarFS.

I'm assuming that an invalid path here would be a fairly exotic thing, and there wouldn't be a great utility in reporting it back to the user. As far as I can tell, the tarfile module also silently ignores these errors, so that swayed me.

I'll wait until somebody asks for it or I can think of an argument against ignoring bad entries...

PR #313, opened 2019-07-24
[MemFS] Release memory when close() is called

### LINKED ISSUE #308: Free memory on MemoryFS.close
Steps to reproduce:

1. Run this piece of code in your favorite REPL.
```python
from fs.memoryfs import MemoryFS

mem_fs = MemoryFS()
with open('large_binary_file', 'rb') as f:
	mem_fs.upload('/large_binary_file', f)
mem_fs.close()
```
2. Watch python process before and after running this script.
3. The memory is there and the files are not removed.
4. Run `del mem_fs` and it immediately claims back the memory.

dargueta:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

MemFS now immediately releases the memory used up by its files when `close()` is called, rather than when the file system gets garbage collected. Closes #308 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.4.9] - (Unreleased)
+
+### Changed
+
+- `MemFS` now immediately releases all memory it holds when `close()` is called,
+  rather than when it gets garbage collected. Closes [issue #308](https://github.com/PyFilesystem/pyfilesystem2/issues/).
+
 ## [2.4.8] - 2019-06-12
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - `MemFS` now immediately releases all memory it holds when `close()` is called,
-  rather than when it gets garbage collected. Closes [issue #308](https://github.com/PyFilesystem/pyfilesystem2/issues/).
+  rather than when it gets garbage collected. Closes [issue #308](https://github.com/PyFilesystem/pyfilesystem2/issues/308).
 
 ## [2.4.8] - 2019-06-12
 
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -207,6 +207,44 @@ class _DirEntry(object):
         if not self.is_dir:
             self._bytes_file = io.BytesIO()
 
+    def clean_up(self, force=False):
+        # type: (bool) -> None
+        """Release resources held by this directory entry.
+
+        Arguments:
+            force (bool):
+                Force remaining open files to close and delete subdirectory entries.
+
+        Raises:
+            fs.errors.ResourceLocked:
+                This directory entry has remaining open files and ``force`` is `False`.
+            fs.errors.DirectoryNotEmpty:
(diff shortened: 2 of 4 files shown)


fs/memoryfs.py
```diff
@@ -207,6 +207,44 @@ def __init__(self, resource_type, name):
...
+
+        Raises:
+            fs.errors.ResourceLocked:
+                This directory entry has remaining open files and ``force`` is `False`.
+            fs.errors.DirectoryNotEmpty:
+                This directory has remaining entries and ``force`` is `False`.
+        """
+        with self.lock:
+            if self._open_files and not force:
+                raise errors.ResourceLocked(
```
willmcgugan: The `close()` method shouldn't throw exceptions. It's not an operation that can fail as such.

fs/memoryfs.py
```diff
@@ -254,7 +292,8 @@ def set_entry(self, name, dir_entry):
 
     def remove_entry(self, name):
         # type: (Text) -> None
-        del self._dir[name]
+        dirent = self._dir.pop(name)
+        dirent.clean_up()
```
willmcgugan: I don't think this is required. If there are no more references to the DirEntry the memory will be garbage collected anyway.

fs/memoryfs.py
```diff
@@ -340,6 +379,11 @@ def _get_dir_entry(self, dir_path):
                 current_entry = current_entry.get_entry(path_component)
             return current_entry
 
+    def close(self):
+        # type: () -> None
+        self.root.clean_up(force=True)
```
willmcgugan: This may be a simpler operation that it appears. If there are no references to `root` then the memory will be garbage collected, and the `clean_up` method won't be required.

Can you try replacing the clean_up with the following:

```
self.root = None
```

or 

```
del self.root
```

tests/test_memoryfs.py
```diff
@@ -1,13 +1,93 @@
...
+
+    def _create_many_files(self):
+        for parent_dir in {"/", "/one", "/one/two", "/one/other-two/three"}:
+            self.fs.makedirs(parent_dir, recreate=True)
+            for file_id in range(50):
+                self.fs.writetext(
+                    posixpath.join(parent_dir, str(file_id)), UNICODE_TEXT
+                )
+
+    @unittest.skipIf(
```
willmcgugan: Thorough test 👍

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.4.9] - (Unreleased)
+
+### Changed
+
+- `MemFS` now immediately releases all memory it holds when `close()` is called,
+  rather than when it gets garbage collected. Closes [issue #308](https://github.com/PyFilesystem/pyfilesystem2/issues/).
```
lurch: Should the URL here be a link to the actual issue?

dargueta: Oops! Good catch.

fs/memoryfs.py
dargueta: Oh yeah, reference counting... :man_facepalming:

dargueta:
I know the contributors list is supposed to be sorted alphabetically but do I *really* need to put my name at the top? Feels weird.

willmcgugan:
@dargueta It's cool. The idea was that the order doesn't indicate significance of contribution.

PR looks good BTW, thanks. I'll give it another review soon.

willmcgugan:
Thanks

mostafa:
Thanks

PR #314, opened 2019-07-26
Allow custom subfs

### LINKED ISSUE #309: Expose SubFS factory
There are some custom methods (i.e. not overrides) on GoogleFS and OneDriveFS to expose specific features of those filesystems.

My problem is that makedir(s) returns a SubFS which then doesn't support those custom methods so you can't use it transparently (e.g. in the tests)
I see that opendir has a factory parameter so that you can customize the return type, but that's missing from the other SubFS-returning functions.
I'd suggest that we could add the same factory parameter to makedir(s), except that it seems to me that it shouldn't be specified by the user but by the filesystem developer - maybe in the _meta dictionary?
I suppose that another way to do it would be to use the custom factory if the user specifies one but if factory is None, use the GoogleFS-specified one.

What do you think?

rkhwaja:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [X] Documentation / docstrings
- [X] Tests
- [ ] Other

## Checklist

## Description

Fixes #309. Allows filesystems to specify a custom type to replace the default SubFS, which is, by default, returned by opendir. Filesystem implementors should return their custom type from makedir(s), but existing filesystems will work unaltered.

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -101,6 +101,9 @@ class FS(object):
     # most FS will use default walking algorithms
     walker_class = Walker
 
+    # default to SubFS
+    subfs_class = None
+
     def __init__(self):
         # type: (...) -> None
         """Create a filesystem. See help(type(self)) for accurate signature.
@@ -1192,7 +1195,7 @@ class FS(object):
         """
         from .subfs import SubFS
 
-        _factory = factory or SubFS
+        _factory = factory or self.subfs_class or SubFS
 
         if not self.getbasic(path).is_dir:
             raise errors.DirectoryExpected(path=path)
--- a/fs/base.py
+++ b/fs/base.py
@@ -1183,7 +1183,7 @@ class FS(object):
             factory (callable, optional): A callable that when invoked
                 with an FS instance and ``path`` will return a new FS object
                 representing the sub-directory contents. If no ``factory``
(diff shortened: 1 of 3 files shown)


rkhwaja:
Unaddressed problems:
- I would have liked to default subfs_class to SubFS but it runs into circular import problems, so it defaults to None instead.
- I didn't change CHANGELOG, because there's no placeholder for the next release.
- I couldn't find the place to add a piece of documentation similar to where walker_class is documented.

fs/base.py
```diff
@@ -1192,7 +1195,7 @@ def opendir(
         """
         from .subfs import SubFS
 
-        _factory = factory or SubFS
+        _factory = factory or self.subfs_class or SubFS
```
althonos: Why not just define `subfs_class = SubFS` directly when declaring the class instead of letting it be `None` ?

rkhwaja: As I said above, it caused circular import dependencies because it means importing SubFS at the top of the module, but SubFS depends on FS.
I'm happy to change it if someone can suggest how it can work. Or y'all can reject it and I'll see if I can work it out myself.

althonos: No it's fine, I had forgotten about that indeed.

willmcgugan:
Thanks.

Issue #33, opened 2017-03-29
Possible bug with MountFS setbinfile() and setbytes()


fresheed:
I try to run following code:
```
(pyfs) [fresheed@localhost pyfs_issue]$ ls -R
.:
temp1  temp2
./temp1:
temp.txt
./temp2:
```
```
(pyfs) [fresheed@localhost pyfs_issue]$ python2.7
...
>>> from fs.osfs import OSFS
>>> f1=OSFS("temp1")
>>> f2=OSFS("temp2")
>>> from fs.mountfs import MountFS
>>> root=MountFS()
>>> root.mount("f1", f1)
>>> root.mount("f2", f2)
>>> root.copy(u"f1/temp.txt", u"f2/temp.txt")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fresheed/.virtualenvs/pyfs/lib/python2.7/site-packages/fs/base.py", line 302, in copy
    self.setbinfile(dst_path, read_file)
  File "/home/fresheed/.virtualenvs/pyfs/lib/python2.7/site-packages/fs/mountfs.py", line 228, in setbinfile
    return fs.setbinfile(path, file)
  File "/home/fresheed/.virtualenvs/pyfs/lib/python2.7/site-packages/fs/base.py", line 977, in setbinfile
    with self.open(path, 'wb') as dst_file:
  File "/home/fresheed/.virtualenvs/pyfs/lib/python2.7/site-packages/fs/osfs.py", line 338, in open
    **options
  File "/home/fresheed/.virtualenvs/pyfs/lib/python2.7/site-packages/fs/error_tools.py", line 77, in __exit__
    traceback
  File "/home/fresheed/.virtualenvs/pyfs/lib/python2.7/site-packages/fs/osfs.py", line 338, in open
    **options
fs.errors.ResourceNotFound: resource 'f2/temp.txt' not found

```

Then I looked in mountfs.py and saw:
```
   def setbinfile(self, path, file):
        self.check()
        fs, _path = self._delegate(path)
        return fs.setbinfile(path, file)
```
So underlying FS receives full path, not delegated _path. Is it a bug or I misunderstood something?
If it's a bug, I can fix it and send you a pull request

fresheed:
Just subclassed MountFS with overrode setbinfile() and setbytes() and got expected behaviour, so it really seems like a bug

willmcgugan:
[...]

PR #330, opened 2019-08-02
Implement TarFS.geturl and ZipFS.geturl and Fix #329, #333, #340 

### LINKED ISSUE #329: Bug: return value of OSFS.geturl cannot be consumed by FS on Windows
`geturl()` will return `file://D:\a\1\s\tests\fixtures\template`(the dog food) on windows, which can be not used(eaten) by osfs itself.

an PR is coming to fix it. I am logging this issue here.

chfw:
## Type of changes

- [x] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Fix #329, #333, #340 

And deliver geturl for ZipFS and TarFS, and my use case is to get url for a file inside a zip, deposit the url, and read it later.

```
        with file_system.open_fs('zip://test.zip') as fs_handle:
                template_file_exists = fs_handle.exists(
                    template
                ) and fs_handle.isfile(template)

                if template_file_exists:
                    return fs_handle.geturl(template)
```

The existing code in ZipFS and TarFS did not imagine an download url is really needed for a file inside a zip or tar. Hence, I would like to present my use case and get it supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed link namespace returning bytes
 - Fixed broken FSURL in windows [#329](https://github.com/PyFilesystem/pyfilesystem2/issues/329)
 - Fixed hidden exception at fs.close() when opening an absent zip/tar file URL [#333](https://github.com/PyFilesystem/pyfilesystem2/issues/333)
+- Fixed abstract class import from `collections` which would break on Python 3.8
+- Fixed incorrect imports of `mock` on Python 3
+- Removed some unused imports and unused `requirements.txt` file
+- Added mypy checks to Travis
+
+### Changed
+
+- Entire test suite has been migrated to [pytest](https://docs.pytest.org/en/latest/). Closes [#327](https://github.com/PyFilesystem/pyfilesystem2/issues/327).
 
 ## [2.4.10] - 2019-07-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,27 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.4.10] - 2019-07-29
+
+### Fixed
+
+- Fixed broken WrapFS.movedir [#322](https://github.com/PyFilesystem/pyfilesystem2/issues/322)
+
+## [2.4.9] - 2019-07-28
+
+### Fixed
+
+- Restored fs.path import
+- Fixed potential race condition in makedirs. Fixes [#310](https://github.com/PyFilesystem/pyfilesystem2/issues/310)
+- Added missing methods to WrapFS. Fixed [#294](https://github.com/PyFilesystem/pyfilesystem2/issues/294)
+
+### Changed
+
+- `MemFS` now immediately releases all memory it holds when `close()` is called,
+  rather than when it gets garbage collected. Closes [issue #308](https://github.com/PyFilesystem/pyfilesystem2/issues/308).
+- `FTPFS` now translates `EOFError` into `RemoteConnectionError`. Closes [#292](https://github.com/PyFilesystem/pyfilesystem2/issues/292)
+- Added automatic close for filesystems that go out of scope. Fixes [#298](https://github.com/PyFilesystem/pyfilesystem2/issues/298)
+
 ## [2.4.8] - 2019-06-12
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [2.4.11] - tba
+## [2.4.11] - Unreleased
 
 ### Added
 
-- Added geturl for TarFS and ZipFS
+- Added geturl for TarFS and ZipFS for 'fs' purpose. NoURL for 'download' purpose.
 
 ### Fixed
 
+- Fixed tests leaving tmp files
+- Fixed typing issues
+- Fixed link namespace returning bytes
 - Fixed broken FSURL in windows [#329](https://github.com/PyFilesystem/pyfilesystem2/issues/329)
 - Fixed hidden exception at fs.close() when opening an absent zip/tar file URL [#33](https://github.com/PyFilesystem/pyfilesystem2/issues/333)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed broken FSURL in windows [#329](https://github.com/PyFilesystem/pyfilesystem2/issues/329)
-
+- Fixed hidden exception at fs.close() when opening an absent zip/tar file URL [#33](https://github.com/PyFilesystem/pyfilesystem2/issues/333)
 
 ## [2.4.10] - 2019-07-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Added geturl for TarFS and ZipFS for 'fs' purpose. NoURL for 'download' purpose.
+- Added helpful root path in CreateFailed exception [#340](https://github.com/PyFilesystem/pyfilesystem2/issues/340)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.4.10] - 2019-07-29
 
+### Added
+
+- Added geturl for TarFS and ZipFS
+
+### Fixed
+
+- Fixed broken FSURL in windows [#329](https://github.com/PyFilesystem/pyfilesystem2/issues/329)
+
+
+## [2.4.10] - 2019-07-29
+
 ### Fixed
 
 - Fixed broken WrapFS.movedir [#322](https://github.com/PyFilesystem/pyfilesystem2/issues/322)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed typing issues
 - Fixed link namespace returning bytes
 - Fixed broken FSURL in windows [#329](https://github.com/PyFilesystem/pyfilesystem2/issues/329)
-- Fixed hidden exception at fs.close() when opening an absent zip/tar file URL [#33](https://github.com/PyFilesystem/pyfilesystem2/issues/333)
+- Fixed hidden exception at fs.close() when opening an absent zip/tar file URL [#333](https://github.com/PyFilesystem/pyfilesystem2/issues/333)
 
 ## [2.4.10] - 2019-07-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [2.4.10] - 2019-07-29
+## [2.4.11] - tba
 
 ### Added
 
--- a/fs/base.py
+++ b/fs/base.py
@@ -310,6 +310,12 @@ class FS(object):
 
         """
 
+    @staticmethod
+    def quote(path_snippet):
+        if six.PY2:
+            path_snippet = path_snippet.encode("utf-8")
+        return six.moves.urllib.parse.quote(path_snippet)
+
     # ---------------------------------------------------------------- #
     # Optional methods                                                 #
     # Filesystems *may* implement these methods.                       #
--- a/fs/base.py
+++ b/fs/base.py
@@ -15,6 +15,7 @@ import os
 import threading
 import time
 import typing
+import platform
 from contextlib import closing
 from functools import partial, wraps
 import warnings
@@ -61,6 +62,7 @@ if False:  # typing.TYPE_CHECKING
 
 
 __all__ = ["FS"]
+_WINDOWS_PLATFORM = platform.system() == "Windows"
 
 
 def _new_name(method, old_name):
@@ -312,9 +314,19 @@ class FS(object):
 
     @staticmethod
     def quote(path_snippet):
-        if six.PY2:
-            path_snippet = path_snippet.encode("utf-8")
-        return six.moves.urllib.parse.quote(path_snippet)
+        if _WINDOWS_PLATFORM and ":" in path_snippet:
+            drive_letter, path = path_snippet.split(":", 1)
+            path = path.replace("\\", "/")
+            if six.PY2:
+                path = path.encode("utf-8")
+            path = six.moves.urllib.parse.quote(path)
+            path_snippet = "{}:{}".format(drive_letter, path)
+        else:
+            path_snippet = path_snippet.replace("\\", "/")
+            if six.PY2:
+                path_snippet = path_snippet.encode("utf-8")
+            path_snippet = six.moves.urllib.parse.quote(path_snippet)
+        return path_snippet
 
     # ---------------------------------------------------------------- #
     # Optional methods                                                 #
--- a/fs/base.py
+++ b/fs/base.py
@@ -8,7 +8,6 @@ can work with any of the supported filesystems.
 
 from __future__ import absolute_import, print_function, unicode_literals
 
-import re
 import abc
 import hashlib
 import itertools
@@ -16,7 +15,6 @@ import os
 import threading
 import time
 import typing
-import platform
 from contextlib import closing
 from functools import partial, wraps
 import warnings
@@ -63,7 +61,6 @@ if False:  # typing.TYPE_CHECKING
 
 
 __all__ = ["FS"]
-_WINDOWS_PLATFORM = platform.system() == "Windows"
 
 
 def _new_name(method, old_name):
@@ -312,20 +309,6 @@ class FS(object):
             >>> my_fs.setinfo('file.txt', details_info)
 
         """
-
-    @staticmethod
-    def quote(path_snippet):
-        if _WINDOWS_PLATFORM and has_drive_letter(path_snippet):
-            drive_letter, path = path_snippet.split(":", 1)
-            if six.PY2:
-                path = path.encode("utf-8")
-            path = six.moves.urllib.request.pathname2url(path)
-            path_snippet = "{}:{}".format(drive_letter, path)
-        else:
-            if six.PY2:
-                path_snippet = path_snippet.encode("utf-8")
-            path_snippet = six.moves.urllib.request.pathname2url(path_snippet)
-        return path_snippet
 
     # ---------------------------------------------------------------- #
     # Optional methods                                                 #
@@ -1662,8 +1645,3 @@ class FS(object):
                     break
                 hash_object.update(chunk)
         return hash_object.hexdigest()
-
-
-def has_drive_letter(path_snippet):
-    windows_drive_pattern = ".:[/\\\\].*$"
-    return re.match(windows_drive_pattern, path_snippet) is not None
--- a/fs/base.py
+++ b/fs/base.py
@@ -317,16 +317,14 @@ class FS(object):
     def quote(path_snippet):
         if _WINDOWS_PLATFORM and has_drive_letter(path_snippet):
             drive_letter, path = path_snippet.split(":", 1)
-            path = path.replace("\\", "/")
             if six.PY2:
                 path = path.encode("utf-8")
-            path = six.moves.urllib.parse.quote(path)
+            path = six.moves.urllib.request.pathname2url(path)
             path_snippet = "{}:{}".format(drive_letter, path)
         else:
-            path_snippet = path_snippet.replace("\\", "/")
             if six.PY2:
                 path_snippet = path_snippet.encode("utf-8")
-            path_snippet = six.moves.urllib.parse.quote(path_snippet)
+            path_snippet = six.moves.urllib.request.pathname2url(path_snippet)
         return path_snippet
 
     # ---------------------------------------------------------------- #
--- a/fs/base.py
+++ b/fs/base.py
@@ -8,6 +8,7 @@ can work with any of the supported filesystems.
 
 from __future__ import absolute_import, print_function, unicode_literals
 
+import re
 import abc
 import hashlib
 import itertools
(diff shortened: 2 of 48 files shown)


chfw:
Hi all, I will get the test cases fixed later

fs/osfs.py
```diff
@@ -588,7 +588,7 @@ def geturl(self, path, purpose="download"):
         # type: (Text, Text) -> Text
         if purpose != "download":
             raise NoURL(path, purpose)
-        return "file://" + self.getsyspath(path)
+        return "file://" + self.getsyspath(path).replace('\\', '/')
```
althonos: I may be wrong, but I'm not sure replacing double-backwards slashes with a single slash is sufficient to have a proper URL, the drive prefix should be URL-encoded as well.

chfw: without the fix, geturl() gives: `file://D:\a\1\s\tests\fixtures\template`. with the fix, it gives `file://D:/a/1/s/tests/fixtures/template`, which fs.open_fs() would happily 'eat'.

tests/test_osfs.py
```diff
@@ -152,3 +151,16 @@ def test_validatepath(self):
...
+        self.fs.create("foo")
+        try:
+            url = self.fs.geturl("foo")
+        except errors.NoURL:
+            self.assertFalse(self.fs.hasurl("foo"))
+        else:
+            self.assertTrue(self.fs.hasurl("foo"))
+
+        # Should not throw an error
+        base_dir = dirname(url)
```
chfw: here is the test for it : @althonos

fs/zipfs.py
```diff
@@ -434,7 +434,8 @@ def removedir(self, path):
     def close(self):
```
chfw: here is relevant doc: about __del__() and what should can be performed in it.

https://docs.python.org/2/reference/datamodel.html#object.__del__

willmcgugan:
@chfw This isn't actually the intent of the `geturl` method. The URL returned was to be something that could be consumed externally by a browser for example. The URLs you have added are FS URLs which may not be of much use outside of PyFilesystem.

Good catch with the missing archive exceptions.

How did you insert those funky icons in the commit messages? Some tool or do you do them manually?

chfw:
@willmcgugan here is the fit commit emoji style guide: https://github.com/slashsbin/styleguide-git-commit-message

I accept that your intent was meant for browser. However, I am using zipfs and tarfs in a new context where: I would need a universal url that pyfs2 willingly read a file out. Both tarfs and zipfs do read content out given: zip://test.zip/my/folder/file. But both refuses to emit url which I find it naturally shall emit. FS has its interface specified already hence all sub fs shall respect where applicable. Then we won’t have to remember which fs can and cannot do.

willmcgugan:
> zip://test.zip/my/folder/file

I'd be surprised if this worked as is. I'd expect that it would consider `test.zip` to be a folder, and it would attempt to open "file" as a zip.

FS URLs do support a syntax that can indicate a path within a filesystem, by separating the filesystem part from the path with a `!`. Your example could be encoded like this:

```
>>> from fs.opener.parse import parse_fs_url
>>> parse_fs_url("zip://test.zip!/my/folder/file")
ParseResult(protocol='zip', username=None, password=None, resource='test.zip', params={}, path='/my/folder/file')
```

The URLs returned must be externally consumable I'm afraid, it could break some projects if we change that. However, there is a `purpose` parameter to `geturl` which could be used to request a FS URL. How about the string `"fs"`?

It would work something like this:

```
>>> my_zip.geturl("bar/egg", purpose="fs")
"zip://foo.zip!bar/egg"
```

And that won't break any existing code as `purpose` defaults to "download"

chfw:
I think your advice is great! My use case is taken care of and the change won't break others.

> "zip://test.zip!/my/folder/file"

Let me update it.

And yes, your doubt is valid. There exists an abstract layer in my lib handles the URL: https://github.com/moremoban/moban/blob/pyfs2/moban/file_system.py#L98

tests/test_osfs.py
lurch: @chfw Based on the discussion in this PR, I guess this `test_consume_geturl` ought to be using the `purpose="fs"` ?

fs/osfs.py
lurch: @chfw I believe @althonos is correct and more work is needed here - see e.g. `https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/` for how to construct "proper" file URIs on Windows

tests/test_osfs.py
chfw: Well, this one is for OSFS. Hence, ‘download’ is assumed.

chfw: This test will ensure the return value of OSFS.geturl() can be consumed by pyfs2 in all os platforms, especially on windows.

lurch: As Will said earlier:
"This isn't actually the intent of the geturl method. The URL returned was to be something that could be consumed externally by a browser for example. The URLs you have added are FS URLs which may not be of much use outside of PyFilesystem."

And as I mentioned in another comment (which GitHub may be auto-hiding) file URLs _are_ a perfectly valid thing (see https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/ ), and may be different to the "FS URLs" which e.g. `open_fs` expects.
Perhaps I wasn't explicit enough, but it looks like perhaps (at least on Windows) `OSFS.geturl()` might be returning "invalid" file URLS? @willmcgugan

chfw: You have made your point. Let me update the code.

tests/test_tarfs.py
```diff
@@ -184,11 +171,34 @@ def test_read_from_filename(self):
...
+            self.fail("Could not close tar fs properly")
+        except Exception:
+            self.fail("Strage exception in closing fs")
+
     def test_getinfo(self):
         super(TestReadTarFS, self).test_getinfo()
         top = self.fs.getinfo("top.txt", ["tar"])
         self.assertTrue(top.get("tar", "is_file"))
 
+    def test_geturl(self):
```
lurch: Perhaps this should be named `test_geturl_for_fs` ?

tests/test_tarfs.py
```diff
@@ -184,11 +171,34 @@ def test_read_from_filename(self):
...
+    def test_read_non_existent_file(self):
+        fs = tarfs.TarFS(open(self._temp_path, "rb"))
+        # it has been very difficult to catch exception in __del__()
+        del fs._tar
+        try:
+            fs.close()
+        except AttributeError:
+            self.fail("Could not close tar fs properly")
+        except Exception:
+            self.fail("Strage exception in closing fs")
```
lurch: typo: Strage -> Strange

tests/test_tarfs.py
```diff
@@ -96,15 +92,6 @@ def destroy_fs(self, fs):
         os.remove(fs._tar_file)
         del fs._tar_file
 
-    def assert_is_bzip(self):
```
lurch: Why was this test deleted? (Looks like it needs to be renamed to `assert_is_gzip`)

chfw: It is because it is not reachable:

Screenshot 2019-08-06 at 18 45 46

`fs` is not imported at all in `test_tarfs.py` so this function shall fail in your unit tests but nothing is failing at all. Hence it proves that it is **dead code**.

chfw: I am wondering why your IDE is hiding this factor from your eyes in past?

fs/osfs.py
chfw: Current change won't support dying IE because `file:///C:/exampleㄓ.txt` -> `file:///C:/example%E3%84%93.txt` by python quote function.

But Microsoft insisted the following for IE browser:
```
Incorrect: file:///C:/example%E3%84%93.txt
Correct: file:///C:/exampleㄓ.txt
```

chfw:
Do you plan to support dead Internet Explorer or not?

If yes, more work is required.

If no, current work is complete from my point of view.

fs/osfs.py
chfw: Here is the reason why IE insists on unicode over percentage quoted strings: https://blogs.msdn.microsoft.com/ieinternals/2014/04/22/unicode-in-url-changes-for-ie11/

But the rest of the world has followed: Berners-Lee's RFC 3986: https://tools.ietf.org/html/rfc3986#page-12

chfw: Here is death ultimatum from Microsoft for IE: https://www.techradar.com/news/microsoft-internet-explorer-dead-1995-to-2019-rip

Definition by 2020, IE die out with python 2.

chfw:
@lurch @willmcgugan , I am going to update my code to delay the fight with Windows File URI. My objective is to get a file url that would work on Windows. But as you can see, I end up in fight with Windows File URI. The whole python community has no answer for it. Definitely, my PR is not intended to solve it. Here is how Windows [solved it](https://referencesource.microsoft.com/#System/net/System/UriExt.cs,1) in 2003 in C#.

The alternative is to use `osfs://`, which is in control of python file system 2 and we can base it on Tim Burner-Lee's rfc. For my PR, I would change the code so that for 'fs' purpose, it will return 'osfs://....' format. And delay the fight for the Windows File URI for future.

fs/osfs.py
willmcgugan: I'm happy to not support IE. Modern browsers should be our target.

chfw: Chrome iterated till version 76 now since 2008.

tests/test_tarfs.py
lurch: IMHO if a test isn't being run (for some reason - I've not looked into the problem) then the "fix" would be to change the unit-tests so the test _does_ get run, not simply delete the test?
ping @willmcgugan ?

> why your IDE is hiding this fact from your eyes in past

Not everyone uses an IDE :wink:  (and I'm sure different IDEs probably "show" different things)

tests/test_osfs.py
```diff
@@ -157,3 +156,42 @@ def test_validatepath(self):
...
+            # raw file, expected url path
+            ["foo", "foo"],
+            ["foo-bar", "foo-bar"],
+            ["foo_bar", "foo_bar"],
+            ["foo/bar ha/barz", "foo/bar%20ha/barz"],
+            ["example b.txt", "example%20b.txt"],
+            ["exampleㄓ.txt", "example%E3%84%93.txt"],
+        ]
+        file_uri_prefix = "osfs://"
+        for test_file, expected_suffix in test_fixtures:
```
lurch: Is `expected_suffix` really the right name for this?

tests/test_osfs.py
```diff
@@ -157,3 +156,42 @@ def test_validatepath(self):
...
+            self.fs.create(test_file)
+            expected = file_uri_prefix + self.fs.getsyspath(expected_suffix).replace(
+                "\\", "/"
+            )
+            try:
+                actual = self.fs.geturl(test_file, purpose="fs")
+            except errors.NoURL:
```
lurch: We probably don't need this `except errors.NoURL` when testing OSFS?

tests/test_osfs.py
```diff
@@ -157,3 +156,42 @@ def test_validatepath(self):
...
+    def test_complex_geturl(self):
+        self.fs.makedirs("foo/bar ha")
+        test_fixtures = [
+            # raw file, expected url path
+            ["foo", "foo"],
+            ["foo-bar", "foo-bar"],
+            ["foo_bar", "foo_bar"],
+            ["foo/bar ha/barz", "foo/bar%20ha/barz"],
+            ["example b.txt", "example%20b.txt"],
+            ["exampleㄓ.txt", "example%E3%84%93.txt"],
```
lurch: Good to see thorough tests :heart:

fs/base.py
```diff
@@ -310,6 +312,22 @@ def setinfo(self, path, info):
 
         """
 
+    @staticmethod
+    def quote(path_snippet):
```
lurch: @willmcgugan Is `base.py` the "right" place for this new `quote` function?
@chfw I wonder if `quote` is too generic a name and something more descriptive would be better? :man_shrugging:

fs/base.py
```diff
@@ -310,6 +312,22 @@ def setinfo(self, path, info):
...
+        if _WINDOWS_PLATFORM and ":" in path_snippet:
+            drive_letter, path = path_snippet.split(":", 1)
+            path = path.replace("\\", "/")
+            if six.PY2:
+                path = path.encode("utf-8")
+            path = six.moves.urllib.parse.quote(path)
+            path_snippet = "{}:{}".format(drive_letter, path)
+        else:
+            path_snippet = path_snippet.replace("\\", "/")
```
lurch: Is this logic correct for non-Windows platforms?
E.g.
```bash
$ touch /tmp/file\\foo:bar
$ ls -l /tmp
-rw-rw-r-- 1 qpc  qpc     0 Aug  8 21:42 file\foo:bar
```

fs/base.py
chfw: @lurch I could find a better place for it. I am happy to put it elsewhere.

url_quote?

tests/test_base.py
```diff
@@ -69,3 +71,18 @@ def mock_getsyspath(path):
...
+
+    def test_quote(self):
+        test_fixtures = [
+            # test_snippet, expected
+            ["foo/bar/egg/foofoo", "foo/bar/egg/foofoo"],
+            ["foo/bar ha/barz", "foo/bar%20ha/barz"],
+            ["example b.txt", "example%20b.txt"],
+            ["exampleㄓ.txt", "example%E3%84%93.txt"],
+        ]
+        if platform.system() == "Windows":
```
lurch: As mentioned above, might be worth also testing paths with `:` and `\` on non-Windows systems?

tests/test_tarfs.py
```diff
@@ -184,11 +171,40 @@ def test_read_from_filename(self):
...
 
+    def test_geturl_for_fs(self):
+        test_fixtures = [
+            # test_file, expected
+            ["foo/bar/egg/foofoo", "foo/bar/egg/foofoo"],
+            ["foo/bar egg/foo foo", "foo/bar%20egg/foo%20foo"],
+        ]
+        for test_file, expected_file in test_fixtures:
+            expected = "tar://{tar_file_path}!/{file_inside_tar}".format(
+                tar_file_path=self._temp_path.replace('\\', '/'),
```
lurch: Only a minor nitpick, but I guess this `self._temp_path.replace('\\', '/')` could be done outside of the for loop?

tests/test_zipfs.py
```diff
@@ -184,8 +209,8 @@ def test_implied(self):
                 z.writestr("foo/bar/baz/egg", b"hello")
             with zipfs.ReadZipFS(path) as zip_fs:
                 foo = zip_fs.getinfo("foo", ["details"])
-                bar = zip_fs.getinfo("foo/bar")
-                baz = zip_fs.getinfo("foo/bar/baz")
+                zip_fs.getinfo("foo/bar")
+                zip_fs.getinfo("foo/bar/baz")
                 self.assertTrue(foo.is_dir)
```
lurch: I wonder if it's worth adding
```python
self.assertTrue(bar.is_dir)
self.assertTrue(baz.is_dir)
```
?

fs/base.py
chfw: No. fixed it.

tests/test_tarfs.py
chfw: Sure! Never nitpick in code review. Updated. Please check it out

tests/test_zipfs.py
chfw: updated with:
```
                self.assertEqual(zip_fs.getinfo("foo/bar").name, "bar")
                self.assertEqual(zip_fs.getinfo("foo/bar/baz").name, "baz")
```

tests/test_base.py
```diff
@@ -79,6 +79,8 @@ def test_quote(self):
             ["foo/bar ha/barz", "foo/bar%20ha/barz"],
             ["example b.txt", "example%20b.txt"],
             ["exampleㄓ.txt", "example%E3%84%93.txt"],
+            ["test/colon:tmp", "test/colon%3Atmp"],
+            ["test/forward\\slash", "test/forward/slash"],
```
lurch: This looks wrong? It means that **both** `"test/forward\\slash"` and `"test/forward/slash"`would get transformed to `"test/forward/slash"` ??

fs/base.py
```diff
@@ -1652,7 +1652,7 @@ def hash(self, path, name):
             fs.errors.UnsupportedHash: If the requested hash is not supported.
 
         """
-        _path = self.validatepath(path)
+        self.validatepath(path)
```
lurch: @willmcgugan does `_path` being an unused variable hint that the `openbin` function below should have been using `_path` instead of `path` ?

tests/test_base.py
chfw: before quoting,  path_snippet = path_snippet.replace("\\", "/") did it.

lurch: Yeah, but surely they should still be different _after_ quoting, otherwise an attempt to access the quoted-version of `"test/forward\\slash"` will actually give you `"test/forward/slash"` which isn't what you want?

```bash
$ echo short > /tmp/file\\foo:bar
$ mkdir /tmp/file
$ echo "really really long" > /tmp/file/foo:bar
$ ls -l /tmp/file\\foo:bar
-rw-rw-r-- 1 qpc qpc 6 Aug  8 22:37 /tmp/file\foo:bar
$ ls -l /tmp/file/foo:bar
-rw-rw-r-- 1 qpc qpc 19 Aug  8 22:37 /tmp/file/foo:bar
```

chfw: `["test/forward\\slash", "test/forward%5Cslash"]` will make you happy

lurch: ...except I always thought that `\` was the backward-slash? :wink:

fs/base.py
chfw: flake8 flag it with red underline.

Screenshot 2019-08-08 at 22 42 31

chfw: bad code never escape

tests/test_base.py
chfw: see the updated commit and I will put it as resolved.

fs/base.py
lurch: @chfw Yeah, but I was wondering if it means that the code _should_ have been doing `self.openbin(_path)` ? (i.e. perhaps it's only an unused variable because of a typo?)

chfw: OK. got you. Let's wait for @willmcgugan

willmcgugan: > @willmcgugan does _path being an unused variable hint that the openbin function below should have been using _path instead of path ?

That's generally the case, but in this instance the return value isn't needed because `openbin` will also call `validatepath`. The right call here would be to discard the return value but keep the call.

If you're wondering why call `validatepath` at all if its called in `openbin`, its so that path errors take priority over the unsupported hash error...

fs/base.py
willmcgugan: I think we should move it out of the base class, wouldn't want to imply it was part of the core interface. Feel free to add a utility file of some sort. Maybe _url_tools.py

@chfw Would you mind adding a docstring and typing comments for this function.

fs/base.py
```diff
@@ -1645,3 +1662,8 @@ def hash(self, path, name):
                     break
                 hash_object.update(chunk)
         return hash_object.hexdigest()
+
+
+def has_drive_letter(path_snippet):
```
willmcgugan: Could we put this at the top of the file and make it private (I'm assuming its not used outside of this file).

Docstring and typing comments please...

lurch:
This PR is getting so big, I wonder if it's worth splitting it into smaller separate PRs? :man_shrugging:

chfw:
Or Merge it now and I will prepare another PR to address your review comments.

chfw:
I have address all review comments now. The ball is in your court.

fs/base.py
chfw: added

fs/base.py
chfw: added.

CHANGELOG.md
```diff
@@ -7,11 +7,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Added geturl for TarFS and ZipFS for 'fs' purpose. NoURL for 'download' purpose.
+- Added helpful root path in CreateFailed exception [#340](https://github.com/PyFilesystem/pyfilesystem2/issues/340)
+
 ### Fixed
 
 - Fixed tests leaving tmp files
 - Fixed typing issues
 - Fixed link namespace returning bytes
+- Fixed broken FSURL in windows [#329](https://github.com/PyFilesystem/pyfilesystem2/issues/329)
+- Fixed hidden exception at fs.close() when opening an absent zip/tar file URL [#33](https://github.com/PyFilesystem/pyfilesystem2/issues/333)
```
lurch: I guess you wanted `[#333]` there instead of `[#33]` ? :wink:

fs/osfs.py
```diff
@@ -584,9 +584,14 @@ def getsyspath(self, path):
 
     def geturl(self, path, purpose="download"):
         # type: (Text, Text) -> Text
-        if purpose != "download":
+        sys_path = self.getsyspath(path)
+        if purpose == "download":
+            return "file://" + self.getsyspath(path)
```
lurch: Should the "download" purpose also be using `url_quote`, e.g. in case the syspath contains a space character?

fs/zipfs.py
```diff
@@ -444,3 +446,12 @@ def readbytes(self, path):
         zip_name = self._path_to_zip_name(path)
         zip_bytes = self._zip.read(zip_name)
         return zip_bytes
+
+    def geturl(self, path, purpose="download"):
+        # type: (Text, Text) -> Text
+        if purpose == "fs" and isinstance(self._file, six.string_types):
+            quoted_file = url_quote(self._file)
+            quoted_path = url_quote(path)
+            return "zip://{}!/{}".format(quoted_file, quoted_path)
```
lurch: Given the commonality between tarfs and zipfs in building this "blah://foo!/bar" path, I wonder if it's worth moving part of this method to `_url_tools.py` ? :man_shrugging:

fs/osfs.py
chfw: addressed your second concern!

to quote the url for "download", I would leave it to #343. I would like to keep the Pandora box closed in this PR. I opened it by accident and regretted it because it will take forever to get this PR merged.

fs/zipfs.py
chfw: Well, marginal gain here. 

Let's do the math: collapse them into one function. (3 + 3 - 2) = 4, meaning 4 less lines. To write a one function into _url_tools.py: 3 lines + 1 function def line + 1 type declaration + 1 docstring + its unit test code lines > 6 lines. In the end, we increase the code base by 2 lines.

lurch: Fair enough, I was just wondering if it was functionality that might be useful to other FSes too in future. But we can always refactor it later if needed.

chfw: Indeed!

chfw:
Could you please approve the PR?

willmcgugan:
@chfw Just want to give it one last look over. Will do that soon.

chfw:
@willmcgugan Any chance for this PR to get released soon?

willmcgugan:
Sorry for the delay @chfw All good, but I've merged a PR and created a merge conflict for you. If you wouldn't mind resolving that and we should be good to go.

chfw:
@willmcgugan , good to merge now.

chfw:
Hooray!

PR #337, opened 2019-08-06
Migrate tests to Pytest

### LINKED ISSUE #327: Switch to pytest?
`nose` is no longer maintained (last release was in 2015) so I think we should consider moving on to a different testing framework. `pytest` gets us a lot of things like simplified argument parametrization and a *lot* of plugins we can install. It works with `unittest` so I don't think the move will require an overhaul.

dargueta:
Closes #327

## Type of changes

- [x] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

* Migrate test suite from `nose` to `pytest`
* Randomized execution order of tests to expose interdependencies
* Overhauled Travis config to use tox environments instead of passing them in multiple places

### Bugs / Cleanup

* Fix a few `import mock` bugs that imported `mock` on Python 3 instead of the native module
* Remove unused imports in a few test files
* Remove `requirements.txt`, as it's out of date and no longer needed.
* Fixed abstract class imports from `collections` which will break on Python 3.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,4 @@ after_success:
   - coveralls
 
 # command to run tests
-script:
-  - tox
+script: tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,25 @@
+dist: xenial
 sudo: false
 language: python
 
+python:
+  - '2.7'
+  - '3.4'
+  - '3.5'
+  - '3.6'
+  - '3.7'
+# TODO (dargueta): Enable these once we figure out SFTP timing issues in PyPy tests
+#  - 'pypy'
+#  - 'pypy3.5'
+
 matrix:
   include:
-    - python: "2.7"
-      env:
-        - SETUPTOOLS=setuptools PIP=pip
-    - python: "3.7"
-      env:
-        - SETUPTOOLS=setuptools PIP=pip
-      dist: xenial
-      sudo: true
-      install:
-        - pip install mypy
-        - make typecheck
-    - python: "3.6"
-      install:
-        - pip install mypy
-        - make typecheck
-      env:
-        - SETUPTOOLS=setuptools PIP=pip
-    - python: "3.5"
-      env:
-        - SETUPTOOLS=setuptools PIP=pip
-    - python: "3.4"
-      env:
-        - SETUPTOOLS=setuptools PIP=pip
+    - name: "Type checking"
+      python: '3.7'
+      env: TOXENV=typecheck
 
 before_install:
-  - pip install $SETUPTOOLS $PIP -U
+  - pip install -U tox tox-travis
   - pip --version
   - pip install -r testrequirements.txt
   - pip freeze
@@ -41,4 +32,4 @@ after_success:
 
 # command to run tests
 script:
-  - nosetests -v --with-coverage --cover-package=fs tests
+  - tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,13 @@ python:
   - '3.5'
   - '3.6'
   - '3.7'
-# TODO (dargueta): Enable these once we figure out SFTP timing issues in PyPy tests
+# TODO (dargueta): Enable these once we figure out FTP timing issues in PyPy tests
 #  - 'pypy'
-#  - 'pypy3.5'
+#  - 'pypy3.5-7.0'
 
 matrix:
   include:
-    - name: "Type checking"
+    - name: 'Type checking'
       python: '3.7'
       env: TOXENV=typecheck
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,0 @@
-appdirs~=1.4.3
-backports.os==0.1.1;  python_version == '2.7'
-enum34==1.1.6 ; python_version < '3.4'
-pytz
-setuptools
-six==1.10.0
-typing==3.6.4 ; python_version < '3.5'
--- a/tests/test_appfs.py
+++ b/tests/test_appfs.py
@@ -1,26 +1,29 @@
 from __future__ import unicode_literals
 
-import unittest
-
+import pytest
 import six
 
-from fs.appfs import UserDataFS
+from fs import appfs
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
 
 
-class TestAppFS(unittest.TestCase):
-    """Test Application FS."""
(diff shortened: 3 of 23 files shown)


dargueta:
Looks like by randomizing the test execution order we've uncovered a bug. `TestOpeners.test_user_data_opener` sometimes crashes in `UserDataFS.__init__` with

```
fs.errors.CreateFailed: root path does not exist
```

See [here](https://travis-ci.org/PyFilesystem/pyfilesystem2/jobs/568144225#L408-L447) for the exact crash.

dargueta:
@willmcgugan do you have any idea what's causing the flakiness in `TestOpeners.test_user_data_opener`? It wasn't happening until I added a pytest plugin to randomize the test execution order so I'm guessing there's a test isolation problem somewhere, but I don't see it.

dargueta:
Looks like I fixed it. Required a bit of fiddling with the fixtures.

willmcgugan:
@dargueta I think I see it. In `test_user_data_opener` the call to `open_fs` doesn't have `create=True`. So it will fail if it hasn't previously been created.

Great work with the pytest migration btw. Much appreciated.

dargueta:
Looks like that fixed it! Thanks!

.travis.yml
```diff
@@ -1,34 +1,25 @@
...
 sudo: false
 language: python
 
+python:
+  - '2.7'
+  - '3.4'
+  - '3.5'
+  - '3.6'
+  - '3.7'
+# TODO (dargueta): Enable these once we figure out SFTP timing issues in PyPy tests
```
lurch: FTP, not SFTP ;-)

tests/test_memoryfs.py
```diff
@@ -14,6 +16,9 @@
     tracemalloc = None
 
 
+@pytest.mark.skipif(
+    not tracemalloc, reason="`tracemalloc` isn't supported on this Python version."
```
lurch: Does moving this up to here now mean the whole of `TestMemoryFS` doesn't get tested if `tracemalloc` isn't available, whereas previously it'd only skip `test_close_mem_free`?

requirements.txt
```diff
@@ -1,7 +0,0 @@
-appdirs~=1.4.3
```
lurch: Why was `requirements.txt` deleted?

lurch:
Is it worth fixing all the `DeprecationWarning: Please use assertEqual instead.` ?

tests/test_memoryfs.py
dargueta: Oooh yeah, good point

requirements.txt
dargueta: It conflicted with requirements in setup.py and isn't used anywhere.

dargueta:
> Is it worth fixing all the DeprecationWarning: Please use assertEqual instead. ?

Those warnings have been present since before the switchover, and in my opinion is kinda outside the scope of this PR (_only_ switch test systems). We can change those as part of a "clean the tests up" follow-up PR.

lurch:
Latest commit is causing failures on both Travis & AppVeyor :confused:

dargueta:
Oops. Dunno how I missed that.

dargueta:
One thing I just noticed is that coverage under pytest says we have 87% coverage, but under nose it says we have 98%. Something doesn't smell right.

willmcgugan:
Is the coverage data being saved across test runs? We want it to be combined so that we include code that only runs on specific environments.

I'd be surprised if there were significant differences in coverage between pytest and node. As far as I know they both use the 'coverage' module under the hood.

requirements.txt
chfw: An advertisement for my project pypi-mobans: requirements of mobanized project are governed centrally by one single source of truth: project-name.yaml. Change in one place, everywhere else follows.

dargueta:
> Is the coverage data being saved across test runs?

Good question. There's a setting for this in coverage or tox somewhere, I'll find it.

requirements.txt
dargueta: Isn't that what `install_requires` in setup.cfg is for?

dargueta:
I [combined coverage](https://github.com/PyFilesystem/pyfilesystem2/pull/337/commits/9d9e50d404941d10b272f6feb87ed957d99e8e63) but it didn't seem to change anything.

tests/test_appfs.py
```diff
@@ -1,26 +1,29 @@
...
-                repr(user_data_fs),
-                "UserDataFS('fstest', author='willmcgugan', version='1.0')",
-            )
-        self.assertEqual(str(user_data_fs), "<userdatafs 'fstest'>")
+from fs import appfs
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
```
chfw: I wondered where `mock` is used.

chfw: here is the search result where `mock` is really used.

https://github.com/PyFilesystem/pyfilesystem2/search?q=mock&unscoped_q=mock

dargueta: Oooh good point -- I took out the last usage and forgot to remove the import too.

dargueta:
@willmcgugan after some digging it appears there is indeed a difference in how coverage is measured. These are the discrepancies:

| File            | Nose  | Pytest |
| ----------      | ----- | ------ |
| _fscompat       | 27    | 0      |
| _version        | 100   | 0      |
| enums           | 100   | 0      |
| errors          | 100   | 0      |
| opener/__init__ | 100   | 0      |
| opener/appfs    | 97    | 0      |
| opener/base     | 90    | 0      |
| opener/errors   | 100   | 0      |
| opener/ftpfs    | 100   | 0      |
| opener/memoryfs | 100   | 0      |
| opener/osfs     | 100   | 0      |
| opener/parse    | 100   | 0      |
| opener/registry | 99    | 0      |
| opener/tarfs    | 100   | 0      |
| opener/tempfs   | 100   | 0      |
| opener/zipfs    | 100   | 0      |
| path            | 100   | 22     |

dargueta:
I think we're good to go now, if no one has any objections?

willmcgugan:
This is great. Thanks!

chfw:
with pytest, the unit test output looks more intuitive:

Screenshot 2019-08-23 at 18 20 21

dargueta:
One of the many benefits of migrating over.

PR #339, opened 2019-08-06
Fix missing ENOTSUP on PyPy

### LINKED ISSUE #338: Bug: PyPy does not have ENOTSUP
To reproduce: Run the unit tests on PyPy. They fail because `errno` doesn't have ENOTSUP defined, even though it's present on the same platform for CPython.

We can fix it by using the CPython value as a default. Error codes aren't guaranteed to be identical across all platforms so this might not work everywhere, but it'll prevent the crash.

dargueta:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

1. Only add `errno.ENOTSUP` to FTP failure codes if that's present.
2. Removed a few unused imports.

Closes #338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed tests leaving tmp files
 - Fixed typing issues
 - Fixed link namespace returning bytes
+- Fixed missing `errno.ENOTSUP` on PyPy. Closes [#338](https://github.com/PyFilesystem/pyfilesystem2/issues/338).
 
 ## [2.4.10] - 2019-07-29
 
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -728,9 +728,8 @@ class FTPFS(FS):
                     for raw_info in self._parse_mlsx(lines):
                         yield Info(raw_info)
                     return
-            with self._lock:
-                for info in self._read_dir(_path).values():
-                    yield info
+            for info in self._read_dir(_path).values():
+                yield info
 
     def scandir(
         self,
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -39,7 +39,6 @@ except ImportError:
         sendfile = None  # type: ignore  # pragma: no cover
 
 from . import errors
-from .errors import FileExists
 from .base import FS
 from .enums import ResourceType
 from ._fscompat import fsencode, fsdecode, fspath
@@ -61,6 +60,7 @@ if False:  # typing.TYPE_CHECKING
         IO,
         List,
         Optional,
+        MutableSet,
         SupportsInt,
         Text,
         Tuple,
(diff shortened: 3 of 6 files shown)


lurch:
I wonder if it might be better to fix this with `try ... catch` rather than adding arbitrary error numbers? :man_shrugging:

lurch:
P.S. Is this at all relevant? https://github.com/giampaolo/pysendfile/issues/28

dargueta:
> I wonder if it might be better to fix this with try ... catch rather than adding arbitrary error numbers?

And what would we do on the catch?

> P.S. Is this at all relevant? giampaolo/pysendfile#28

I don't think so?

dargueta:
Crashes in PyPy appear to be unrelated, due to an SFTP connection issue.

fs/osfs.py
```diff
@@ -420,17 +420,18 @@ def _check_copy(self, src_path, dst_path, overwrite=False):
...
+            errno.EINVAL,
+            errno.ENOSYS,
+            errno.EBADF,
+            errno.ENOTSOCK,
+            errno.EOPNOTSUPP,
+        }  # type: MutableSet[int]
+
+        # PyPy doesn't define ENOTSUP so we have to add it conditionally.
+        if hasattr(errno, "ENOTSUP"):
+            _sendfile_error_codes.add(errno.ENOTSUP)
```
lurch: I guess you _could_ convert `_sendfile_error_codes` back into a `frozenset` after adding (or not adding) the `errno.ENOTSUP` ?

dargueta: Does it need to be one?

lurch: Dunno - but I guess it doesn't hurt? I'll leave it up to @willmcgugan to decide :slightly_smiling_face:

fs/osfs.py
```diff
@@ -420,17 +420,18 @@ def _check_copy(self, src_path, dst_path, overwrite=False):
...
-            }
-        )
+        _sendfile_error_codes = {
+            errno.EIO,
+            errno.EINVAL,
+            errno.ENOSYS,
+            errno.EBADF,
+            errno.ENOTSOCK,
+            errno.EOPNOTSUPP,
+        }  # type: MutableSet[int]
```
willmcgugan: I think the type is redundant here. Mypy can deduce the types of literals.

fs/osfs.py
willmcgugan: A regular set will do the job. Don't see any scope for breakage.

fs/osfs.py
lurch: Does that mean the change on line 63 is also redundant?

dargueta: Probably

willmcgugan:
The pypy build is failing on the ftpfs tests. I've seen this before and never figured it out. It's unrelated to your changes.

I'm happy to merge this and treat the pypy test fails as another task. What do you think?

dargueta:
> I'm happy to merge this and treat the pypy test fails as another task.

Conceptually it makes sense, but anyone creating a new PR is gonna be really mad that their PR is broken because of something I did. Up to you.

willmcgugan:
How about remove the pypy builds from travis for now, and I'll add an issue to add them later?

willmcgugan:
Added an issue. https://github.com/PyFilesystem/pyfilesystem2/issues/342

If you wouldn't mind removing the pypy builds I'll merge this...

dargueta:
This is gonna have a merge conflict with the pytest PR #337 which modifies the Travis file and will remove the PyPy tests. Let's get that in first and then this will be good to go.

willmcgugan:
@dargueta Would you mind fixing that merge conflict and we should be good to go.

fs/ftpfs.py
```diff
@@ -728,9 +728,8 @@ def _scandir(self, path, namespaces=None):
                     for raw_info in self._parse_mlsx(lines):
                         yield Info(raw_info)
                     return
-            with self._lock:
-                for info in self._read_dir(_path).values():
-                    yield info
+            for info in self._read_dir(_path).values():
+                yield info
```
lurch: Does this change have any impact on non-pypy usages of pyfilesystem?

willmcgugan: I think this was purely because the lock is redundant, as there is a lock on L714.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed incorrect imports of `mock` on Python 3
 - Removed some unused imports and unused `requirements.txt` file
 - Added mypy checks to Travis
+- Fixed missing `errno.ENOTSUP` on PyPy. Closes [#338](https://github.com/PyFilesystem/pyfilesystem2/issues/338).
```
lurch: Will this PR also close #342 ?

willmcgugan:
Thanks @dargueta !

Issue #332, opened 2019-08-02
Enforce mypy typechecking in CI


chfw:
Hi all,

This project's PR template reads very interesting:

` I accept that @willmcgugan may be pedantic in the code review.`

I wondered if adding flake8 check to such a cool project would support the word: 'pedantic', strongly. Is there a good reason not to have flake8?

Here is what flake8 tells about the files:
```
...
tests/test_path.py:183:22: F405 'iswildcard' may be undefined, or defined from star imports: fs.path
tests/test_path.py:184:22: F405 'iswildcard' may be undefined, or defined from star imports: fs.path
tests/test_path.py:185:22: F405 'iswildcard' may be undefined, or defined from star imports: fs.path
tests/test_path.py:186:26: F405 'iswildcard' may be undefined, or defined from star imports: fs.path
tests/test_path.py:187:26: F405 'iswildcard' may be undefined, or defined from star imports: fs.path
tests/test_path.py:188:26: F405 'iswildcard' may be undefined, or defined from star imports: fs.path
tests/test_path.py:204:30: F405 'relativefrom' may be undefined, or defined from star imports: fs.path
...
```

willmcgugan:
I've found flake8 to have too may false positives as it only has a superficial understanding of the code. We do have typechecking in the project with mypy, which is far better at detecting issues in my experience. For instance mypy would know the difference between a star import and undefined symbol.

We don't have the mypy check in the CI though, and we really should. A few mypy errors have crept in since I last checked, so that needs to be addressed.

@chfw I'm going to accept this issue, but change the title...

willmcgugan:
Mypy issues fixed in master.

chfw:
Thanks for accepting my issue. Yes, mypy definitely will help enforce the type check and help reducing runtime erros. 👍 

With your attention, please allow me to explain a bit on the positive side of flake8. I am pleased already and I am not going to push like a nut :). That's only me so I would not like to push everyone.

My emacs or pycharm will flag up lines to indicate potential problems. Normally, you would like to have no red lines at a glance. Take tools.py as an example, the IDE tells it has some problems: imported but not used.

![Screenshot 2019-08-04 at 20 25 00](https://user-images.githubusercontent.com/4280312/62428167-7012c880-b6f6-11e9-8cc5-47a2098ecb11.png)

Yet again, let's stress that my thanks for accepting my feedback. I am pleased that you took further action. It is appreciated!

willmcgugan:
[...]

PR #351, opened 2019-08-28
Add flake8 linting

dargueta:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [x] Other

## Checklist

### Bugs (or potential ones)

* Fixed `getattr` call that would cause an AttributeError where it was supposed to prevent it.
* Replaced overbroad except when serializing JSON with its documented exceptions. This would otherwise have hidden problems we weren't expecting.
* Replaced bare `except:` with `except Exception:`, see [here](https://docs.python.org/2/howto/doanddont.html#except) for why.

## Cleanup

* Removed unused imports
* Fixed some long lines
* Explicitly marked unused variables with leading underscore
* Removed unused variables
* Replaced `if False` with `if typing.TYPE_CHECKING`
* Replaced `from *` with explicit imports in some tests

## Other

* Added flake8 for linting
* Minor tweak to coverage so `@overload` is also ignored, not just `@typing.overload`

Expansion on discussion in #332.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ matrix:
     - name: 'Type checking'
       python: '3.7'
       env: TOXENV=typecheck
+    - name: 'Lint'
+      python: '3.7'
+      env: TOXENV=lint
 
 before_install:
   - pip install -U tox tox-travis
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,11 +22,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed abstract class import from `collections` which would break on Python 3.8
 - Fixed incorrect imports of `mock` on Python 3
 - Removed some unused imports and unused `requirements.txt` file
-- Added mypy checks to Travis
+- Added mypy checks to Travis. Closes [#332](https://github.com/PyFilesystem/pyfilesystem2/issues/332).
+- Fixed bug in a decorator that would trigger an `AttributeError` when a class
+  was created that implemented a deprecated method and had no docstring of its
+  own.
 
 ### Changed
 
 - Entire test suite has been migrated to [pytest](https://docs.pytest.org/en/latest/). Closes [#327](https://github.com/PyFilesystem/pyfilesystem2/issues/327).
+- Style checking is now enforced using `flake8`; this involved some code cleanup
+  such as removing unused imports.
 
 ## [2.4.10] - 2019-07-29
 
--- a/fs/_repr.py
+++ b/fs/_repr.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 
 import typing
 
-if False:  # typing.TYPE_CHECKING
+if typing.TYPE_CHECKING:
     from typing import Text, Tuple
 
 
--- a/fs/base.py
+++ b/fs/base.py
@@ -1624,7 +1624,8 @@ class FS(object):
 
         Arguments:
             path(str): A path on the filesystem.
-            name(str): One of the algorithms supported by the hashlib module, e.g. `"md5"`
+            name(str):
+                One of the algorithms supported by the hashlib module, e.g. `"md5"`
 
         Returns:
             str: The hex digest of the hash.
--- a/fs/base.py
+++ b/fs/base.py
@@ -84,7 +84,7 @@ def _new_name(method, old_name):
 """.format(
         method.__name__
     )
-    if getattr(_method, "__doc__"):
+    if hasattr(_method, "__doc__"):
         _method.__doc__ += deprecated_msg
 
     return _method
--- a/fs/base.py
+++ b/fs/base.py
@@ -28,7 +28,7 @@ from .path import abspath, join, normpath
 from .time import datetime_to_epoch
 from .walk import Walker
 
-if False:  # typing.TYPE_CHECKING
+if typing.TYPE_CHECKING:
     from datetime import datetime
     from threading import RLock
     from typing import (
--- a/fs/filesize.py
+++ b/fs/filesize.py
@@ -16,7 +16,7 @@ from __future__ import unicode_literals
 
 import typing
 
-if False:  # typing.TYPE_CHECKING
+if typing.TYPE_CHECKING:
     from typing import Iterable, SupportsInt, Text
 
 
--- a/fs/filesize.py
+++ b/fs/filesize.py
@@ -34,7 +34,8 @@ def _to_str(size, suffixes, base):
     elif size < base:
         return "{:,} bytes".format(size)
 
-    for i, suffix in enumerate(suffixes, 2):
+    # TODO (dargueta): Don't rely on unit or suffix being defined in the loop.
+    for i, suffix in enumerate(suffixes, 2):  # noqa: B007
         unit = base ** i
         if size < unit:
             break
--- a/fs/info.py
+++ b/fs/info.py
@@ -69,16 +69,16 @@ class Info(object):
         return self.raw == getattr(other, "raw", None)
 
     @overload
-    def _make_datetime(self, t):  # pragma: no cover
+    def _make_datetime(self, t):
         # type: (None) -> None
         pass
 
-    @overload
-    def _make_datetime(self, t):  # pragma: no cover
+    @overload  # noqa: F811
+    def _make_datetime(self, t):
         # type: (int) -> datetime
         pass
 
-    def _make_datetime(self, t):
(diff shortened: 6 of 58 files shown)


CHANGELOG.md
```diff
@@ -22,11 +22,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Added mypy checks to Travis. Closes [#332](https://github.com/PyFilesystem/pyfilesystem2/issues/332).
+- Fixed bug in a decorator that would trigger an `AttributeError` when a class
+  was created that implemented a deprecated method and had no docstring of its
+  own.
 
 ### Changed
 
 - Entire test suite has been migrated to [pytest](https://docs.pytest.org/en/latest/). Closes [#327](https://github.com/PyFilesystem/pyfilesystem2/issues/327).
+- Style checking is now enforced using `flake8`; this involved some code cleanup
+  such as removing unused imports.
```
chfw: 👍

.travis.yml
```diff
@@ -17,6 +17,9 @@ matrix:
     - name: 'Type checking'
       python: '3.7'
       env: TOXENV=typecheck
+    - name: 'Lint'
```
chfw: Once and forever

fs/filesize.py
```diff
@@ -34,7 +34,8 @@ def _to_str(size, suffixes, base):
     elif size < base:
         return "{:,} bytes".format(size)
 
-    for i, suffix in enumerate(suffixes, 2):
+    # TODO (dargueta): Don't rely on unit or suffix being defined in the loop.
```
willmcgugan: As long as there is at least one value in `suffixes` we can rely on those variables being defined. Perhaps an `assert len(suffixes)` is appropriate?

fs/base.py
```diff
@@ -84,7 +84,7 @@ def _method(*args, **kwargs):
 """.format(
         method.__name__
     )
-    if getattr(_method, "__doc__"):
+    if hasattr(_method, "__doc__"):
```
willmcgugan: Good catch. I guess this has always been True. Did flake8 flag this?

tests/test_path.py
```diff
@@ -8,7 +8,29 @@
 
 import unittest
 
-from fs.path import *
+from fs.path import (
```
chfw: 👍

fs/info.py
```diff
@@ -69,33 +69,33 @@ def __eq__(self, other):
         return self.raw == getattr(other, "raw", None)
 
     @overload
-    def _make_datetime(self, t):  # pragma: no cover
+    def _make_datetime(self, t):
```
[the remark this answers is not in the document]
dargueta: Adding `@overload` to the coverage settings removed the need for it.

fs/test.py
```diff
@@ -485,8 +485,8 @@ def test_getinfo(self):
         # Raw info should be serializable
         try:
             json.dumps(info)
-        except:
-            assert False, "info should be JSON serializable"
+        except (TypeError, ValueError):
```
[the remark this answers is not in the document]
dargueta: I'd argue that's a bug in user code and not that it's not JSON-serializable. By hiding the 1/0 problem we're doing the programmer a disservice, no?

fs/base.py
dargueta: Yeah. `getattr(thing, "some_attribute")` is no safer than `thing.some_attribute` without passing a default as the third argument since it'll still raise an `AttributeError`. I figured `hasattr` was the intent.

fs/filesize.py
dargueta: That'd work.

fs/info.py
```diff
@@ -69,33 +69,33 @@ def __eq__(self, other):
...
         # type: (None) -> None
         pass
 
-    @overload
-    def _make_datetime(self, t):  # pragma: no cover
+    @overload  # noqa: F811
```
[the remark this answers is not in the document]
dargueta: Right. A fix has already been pushed to the flake8 master for it but that was super recent (last couple weeks) so it hasn't been released yet. Once that new release comes out we can remove these.

fs/_repr.py
```diff
@@ -5,7 +5,7 @@
 
 import typing
 
-if False:  # typing.TYPE_CHECKING
+if typing.TYPE_CHECKING:
```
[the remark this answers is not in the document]
dargueta: Yep, wasn't added until 3.5.2. We're on 3.6 now so it's no longer an issue.

dargueta:
> In my experience flake8 can be a neutral or even negative influence on code quality, if you have good code coverage.

Interesting! I'm curious to hear more.

dargueta:
Oh also -- for the openers, is there a reason why we import the file system classes *inside* the opener function? Circular dependencies or something?

fs/filesize.py
dargueta: So actually I found an edge case -- this will break for values above 10<sup>24</sup> bytes (which is absurdly high) because it'll run off the edge of the array and `unit` will never be defined.

While incredibly unlikely to be encountered in practice, a troll could still use this to break things. I've modified it to both get rid of the linter warning *and* behave correctly in the event this gets handed a huge number.

fs/test.py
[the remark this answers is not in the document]
willmcgugan: This is a unit test so we wouldn't be hiding it, we'd be emphasizing it. In practice the only difference is an error becomes a fail, but at least we'd give a clearer message in the assertion?

fs/base.py
willmcgugan: Yeah that was the intent, just curious what the flake8 error was. I guess it noticed that the attribute name was a literal.

fs/filesize.py
willmcgugan: I don't see how it will break. It will use the last unit, which may result in an absurdly large value, bit it's still valid.

```python
>>> from fs.filesize import traditional
>>> traditional(1E32)
'82,718,061.3 YB'
>>> traditional(1E64)
'8,271,806,125,530,276,925,072,459,481,976,663,965,696.0 YB'
```

willmcgugan:
@dargueta It may be a cultural thing, but orgs I've worked for wanted to stick ruthlessly to flake8 and discouraged exceptions of any kind. Which meant that sometimes you modify good code for zero benefit (possibly even making it worse). But I could come round to a more pragmatic use of flake8.

re the openers. Memory is a bit hazy, but I think that was more to do with reducing the time to do `import fs`. Lazily importing the filesystems means they don't need to be imported when the openers are registered.

chfw:
@willmcgugan I could see that you are not convinced whether flake8 would result in good code. As I mentioned about unused imports, dead variables in other issues, they are caught at design time when flake8 is integrated with IDE. Each line that flake8 flags up indicates some underlying issue. Hence, when an alien python file is opened, one would automatically look for red lines. All these issues are very cheap to identify.

> Which meant that sometimes you modify good code for zero benefit (possibly even making it worse).

Well, I think this is not related to flake8 introduction. Any other code change may result in unwanted changes. I would argue why unit tests, integration tests would fail to warn the new developer who would break  good code. The fear that unwanted code change, which usually is due to code refactoring, will break good code exists in all software projects, is affecting all software developers and dominates many code review discussions. Unless we face it, it will hunt us forever. My advice is to fight with it with more tests, whichever reproducible way to test our fear.

Then back to flak8 discussion, because it is cheap to find issues at design time, it is also cheap to get ride of at design time with TDD. And we keep such effort on daily basis so that the code will not smell like a still pond.

fs/filesize.py
dargueta: Odd...broke for me. o.O

dargueta:
> wanted to stick ruthlessly to flake8 and discouraged exceptions of any kind

Oh yeah, I see how that'd be a problem. I am all up for disabling warnings when they interfere with productivity and readability, but warnings about bare `except` statements, unused imports, etc. are valuable. For example, an unused import could be the cause of a circular dependency that we end up designing around when the original code would've worked without that unused import.

fs/filesize.py
dargueta: I removed the change btw

fs/test.py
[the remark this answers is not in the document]
dargueta: Well we'd be hiding the original stack trace on Python 2 and probably making it harder to read on Python 3, no?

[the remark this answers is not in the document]
willmcgugan: Ok, lets go with this.

willmcgugan:
Thanks @dargueta !

PR #354, opened 2019-10-04
Use `unittest.SkipTest` instead of `pytest.skip` in `fs.test`

althonos:
## Type of changes

- [x] Bug fix

## Description

Hi Will!

Just a quick bugfix since switching to `pytest` broke the CI of `fs.sshfs` and `fs.smbfs`, as I am not using `pytest` for these :smiley: 

Instead of using `pytest.skip`, raising `unittest.SkipTest` will be recognized in both `pytest` and non-`pytest` test harnesses, since `unittest` is pretty much the standard as it is part of the standard library.

### DIFF
--- a/fs/test.py
+++ b/fs/test.py
@@ -15,8 +15,7 @@ import json
 import math
 import os
 import time
-
-import pytest
+import unittest
 
 import fs.copy
 import fs.move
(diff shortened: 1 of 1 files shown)


willmcgugan:
Thanks @althonos

lurch:
@willmcgugan Not related to this PR directly, but Travis seems to be failing on "Type checking" ?

willmcgugan:
Yeah, noticed that. Not sure why that popped up since it was passing previously. Will take a look at some point.

Issue #342, opened 2019-08-08
FTP tests fail on Pypy


willmcgugan:
FTPFS tests fail on Pypy. We should fix that and add Pypy builds to Travis.

chfw:
here is the travis snippet that you need:

```
python:
  - &pypy2 pypy2.7-6.0
  - &pypy3 pypy3.5-6.0
```

reference:

https://github.com/moremoban/pypi-mobans/blob/dev/templates/travis.yml.jj2

dargueta:
Was there a bug introduced in PyPy 7.x?

dargueta:
Considering Python 2 has been EOL for months now I think we can ignore the PyPy 2 CI failures and officially declare PyPy support.

PR #357, opened 2019-10-18
Start testing PyPy

dargueta:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

There's a problem with the SFTP server daemon in our tests that causes PyPy to fail on them (see #342). We can still test PyPy compatibility with other parts of the codebase without failing the build by using Travis' [`allow_failures`](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail) directive. This will still run tests on PyPy, but won't fail the build if the PyPy builds don't pass. We can remove the `allow_failures` part once #342 is fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,8 @@ python:
   - "3.6"
   - "3.7"
   - "3.8-dev"
-# TODO (dargueta): Enable these once we figure out FTP timing issues in PyPy tests
-#  - 'pypy'
-#  - 'pypy3.5-7.0'
+  - 'pypy'
+  - 'pypy3.5-7.0'  # Need 7.0+ due to a bug in earlier versions that broke our tests.
 
 matrix:
   include:
@@ -21,6 +20,11 @@ matrix:
     - name: 'Lint'
       python: '3.7'
       env: TOXENV=lint
+
+  # Temporary bandaid for https://github.com/PyFilesystem/pyfilesystem2/issues/342
+  allow_failures:
+    - python: pypy
+    - python: pypy3.5-7.0
 
 before_install:
   - pip install -U tox tox-travis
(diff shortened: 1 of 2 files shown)


dargueta:
@willmcgugan do you know anything about the MyPy failures?

willmcgugan:
No, they appeared recently. It's on my list to take a look at.

I'm traveling atm. Back next week.

althonos:
seems like the recent version wants `X.__new__` to return a instance of `X`, but that's incompatible with the factory pattern we are using in `ZipFS` and `TarFS`.

dargueta:
> seems like the recent version of mypy

Should we pin it to an earlier version for now?

althonos:
Dissecting the version show that the new build fails with `mypy == 0.740` but succeeds with `mypy == 0.720`.  Incriminated PR is likely https://github.com/python/mypy/pull/7656.

willmcgugan:
I think I'd prefer to ignore those errors rather than pin to older Mypy.

The factory pattern is a little unorthodox, but I wouldn't say it's a bug.

althonos:
@willmcgugan : same, this can be ignored IMO.

lurch:
I know nothing about MyPy, but is it worth reporting a bug / feature-request asking them to support the "unorthodox factory pattern" being used in PyFilesystem? Or is a "local ignore" the only sensible way around this?

willmcgugan:
I reckon 99% of the time that would be a bug. So Mypy is right to flag it,  but we have a legit reason to ignore those errors.

I'll review that code though. Maybe there is a way of making it less unorthodox that Mypy approves of.

dargueta:
> I think I'd prefer to ignore those errors rather than pin to older Mypy.

I forgot we can do that 😄

willmcgugan:
Thanks @dargueta

Issue #360, opened 2019-12-03
Way to open password protected zip file


tzing:
As python built-in `zipfile` provides [setpassowrd](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.setpassword) for opening the password protected zip file,
is it proper to add a interface to set password on `ReadZipFS` and even `ZipFS`?

I known there is a workaround is to access the private member
```py
zipf._zip.setpassword( ... )
```
but i still want a more convenience way

Yet another problem is-- `setpassword` is only valid for extract file but not for compression.
Just wonder if it would be the concern on creating this sort of interface.

lurch:
Yeah, probably makes sense to add a `pwd=None` parameter to the ZipFs constructor, and then raise an error if `write == True and pwd is not None` ?

Looking at Python's zipfile docs, it seems there's a 'default password' for the zipfile (set with `setpassword`) and also the option to have different passwords for each file within the zip https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.read
I didn't even know that per-file passwords were something that zip files allowed :wink: 

I guess ZipFs could support the default passsword, but there's probably no way for it to support per-file passwords.

willmcgugan:
> Yeah, probably makes sense to add a pwd=None parameter to the ZipFs constructor, and then raise an error if write == True and pwd is not None ?

👍 

> I guess ZipFs could support the default password, but there's probably no way for it to support per-file passwords.

It would be possible to implement per-file passwords with the `options` keyword arguments on `open` and other methods.

I don't know if anything has changed regarding passwords in zip, but my understanding was that the password protection was very weak and vulnerable to brute force cracking. Not opposed to adding zip password support though, if it is useful to someone.

PR #361, opened 2019-12-06
password support to ZipFS

tzing:
## Type of changes

- [ ] Bug fix
- [v] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [v] I've run the latest [black](https://github.com/ambv/black) with default args on new code.
- [v] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.
- [v] I've added tests for new code.
- [v] I accept that @willmcgugan may be pedantic in the code review.

## Description

Related issue: #360 

This PR add `passwd` argument for `ZipFS` and `ReadZipFS` constructor to set the default password for extracting the file. And for per-file password, there is also `passwd` arg on `openbin` and `readbytes` to override the default settings. 

Extra, I added `setpassword` on `ReadZipFS` since users could get the fs from `open_fs` and there would be no place the set password.

One thing to be mention, it changes `readbytes`'s signature: 
```diff
-    def readbytes(self, path):
-        # type: (Text) -> bytes
+    def readbytes(self, path, passwd=None):
+        # type: (Text, Optional[AnyStr]) -> bytes
```
and thus it does not follow the form on the base class. 
Please let me know if you have any concern.

### DIFF
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -483,6 +483,4 @@ class ReadZipFS(FS):
         # type: (AnyStr) -> None
         """Set *passwd* as default password to extract encrypted files.
         """
-        if passwd is None:
-            raise ValueError('passwd')
         self._zip.setpassword(_bytes(passwd))
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -461,13 +461,13 @@ class ReadZipFS(FS):
         if hasattr(self, "_zip"):
             self._zip.close()
 
-    def readbytes(self, path, pwd=None):
+    def readbytes(self, path, passwd=None):
         # type: (Text, Optional[AnyStr]) -> bytes
         self.check()
         if not self._directory.isfile(path):
             raise errors.ResourceNotFound(path)
         zip_name = self._path_to_zip_name(path)
-        zip_bytes = self._zip.read(zip_name, pwd=_bytes(pwd))
+        zip_bytes = self._zip.read(zip_name, pwd=_bytes(passwd))
         return zip_bytes
 
     def geturl(self, path, purpose="download"):
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -43,13 +43,27 @@ if typing.TYPE_CHECKING:
     R = typing.TypeVar("R", bound="ReadZipFS")
 
 
+def _bytes(s):
+    # type: (AnyStr) -> bytes
+    if s is None:
+        return None
+    elif isinstance(s, six.binary_type):
+        return s
+    elif isinstance(s, six.string_types):
+        return s.encode()
+    else:
+        raise TypeError("expected string type or byte type, not " + type(s).__name__)
+
+
 class _ZipExtFile(RawWrapper):
-    def __init__(self, fs, name):
-        # type: (ReadZipFS, Text) -> None
+    def __init__(self, fs, name, passwd=None):
+        # type: (ReadZipFS, Text, Optional[AnyStr]) -> None
         self._zip = _zip = fs._zip
         self._end = _zip.getinfo(name).file_size
         self._pos = 0
-        super(_ZipExtFile, self).__init__(_zip.open(name), "r", name)
+        super(_ZipExtFile, self).__init__(
+            _zip.open(name, pwd=_bytes(passwd)), "r", name
+        )
 
     def read(self, size=-1):
         # type: (int) -> bytes
@@ -185,7 +199,7 @@ class ZipFS(WrapFS):
                 file, compression=compression, encoding=encoding, temp_fs=temp_fs
             )
         else:
-            return ReadZipFS(file, encoding=encoding)
+            return ReadZipFS(file, encoding=encoding, passwd=passwd)
 
     if typing.TYPE_CHECKING:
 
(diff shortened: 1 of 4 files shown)


willmcgugan:
I'm afraid the changed signature is a deal breaker. There is an `options` keyword args on the `open` methods which I was thinking could be used for an optional `password` argument. That should probably be added to all shortcut methods that read or write data.

So the base method could be:

```
    def readbytes(self, path, **options):
        # type: (Text) -> bytes
        with closing(self.open(path, mode="rb", **options)) as read_file:
            contents = read_file.read()
        return contents
```

That would allow ZipFS to use a per-file password, while keeping the same signature. There would need to be a similar change to `readtext`, `writebytes`, and `writetext`.  And it would have to be changed everywhere, not just the base.

Sorry this may be a large change that it appeared.

> Extra, I added setpassword on ReadZipFS since users could get the fs from open_fs and there would be no place the set password.

FS urls support [parameters](https://docs.pyfilesystem.org/en/latest/openers.html#url-parameters) which you could use to pass in the password.

fs/zipfs.py
```diff
@@ -290,13 +311,15 @@ class ReadZipFS(FS):
     }
 
     @errors.CreateFailed.catch_all
-    def __init__(self, file, encoding="utf-8"):
-        # type: (Union[BinaryIO, Text], Text) -> None
+    def __init__(self, file, encoding="utf-8", passwd=None):
```
willmcgugan: Can you rename `passwd` to `password`. I prefer to avoid abbreviations where possible.

fs/zipfs.py
```diff
@@ -43,13 +44,27 @@
     R = typing.TypeVar("R", bound="ReadZipFS")
 
 
+def _bytes(s):
```
willmcgugan: Not keen on this dynamic handling of bytes/str. I know it seems like a convenience, but it tends to hide bugs. Better to check what format the Python zipfile library accepts and stick with that. If its different on Python versions, you may need an `if PY3` branch.

tzing: Both py2 and py3 built-in `zipfile` lib only supports bytes for password but bytes is not intuitive for humans.
If this conversion is not accepted, I could and a type check and raise error if user not provides bytes password.

willmcgugan: > If this conversion is not accepted, I could and a type check and raise error if user not provides bytes password.

:+1:

The URL parameters would have to be an exception as they are always unicode/str.

tzing:
I'm a little bit confusing.

Just to be checked - there's two things I could do:
1. remove `_bytes` and `setpassword` would only allow bytes, since python's zipfile lib only accept bytes
2. add URL parameter support for opening password protected zip, and there would have a auto conversion from str to bytes since URL parameters is always str.

But what about `readbytes`? If you mean it is needed to change the base class's signature, it seems shouldn't be done in this PR since it is for zip file? or I should restore the signature temporary?

Thanks for your kindly reply.

willmcgugan:
> remove _bytes and setpassword would only allow bytes, since python's zipfile lib only accept bytes

:+1:

> add URL parameter support for opening password protected zip, and there would have a auto conversion from str to bytes since URL parameters is always str.

:+1:

> But what about readbytes? If you mean it is needed to change the base class's signature, it seems shouldn't be done in this PR since it is for zip file? or I should restore the signature temporary?

If a signature changes, it needs to be changed everywhere. i.e. the base, ZipFS and the other built in filesystems. One of the goals of PyFilesystem is that you can swap one filesystem for another, and your code will still work.

The base and other filesystems can ignore the password, so it would just be a null-op for them.

On reflection, the argument should probably be called `zip_password` to avoid any future name collision.

lurch:
>> But what about readbytes? If you mean it is needed to change the base class's signature, it seems shouldn't be done in this PR since it is for zip file?

> If a signature changes, it needs to be changed everywhere. i.e. the base, ZipFS and the other built in filesystems. One of the goals of PyFilesystem is that you can swap one filesystem for another, and your code will still work.

It's possibly worth doing that in a separate PR, and once that's merged, then come back to this password-protected-zip PR? :man_shrugging:

lurch:
> On reflection, the argument should probably be called zip_password to avoid any future name collision.

Did you mean "dictionary key" rather than "argument" ?

willmcgugan:
> It's possibly worth doing that in a separate PR, and once that's merged, then come back to this password-protected-zip PR? 🤷‍♂

Yeah, nice to have one feature per PR. 

> Did you mean "dictionary key" rather than "argument" ?

The options are collected kwargs, so I guess that would make it an argument in the call and a dictionary key in the method.

lurch:
> The options are collected kwargs, so I guess that would make it an argument in the call and a dictionary key in the method.

Ahh, I'd missed that you were using `options` to collect the kwargs - I naively (without looking at the code) assumed you were passing in an `options` dictionary.

fs/zipfs.py
```diff
@@ -456,3 +468,16 @@ def geturl(self, path, purpose="download"):
...
+        """Set *password* as default password to extract encrypted files.
+        """
+        _password_type_check(password)
+        self._zip.setpassword(password)
+
+
+def _password_type_check(password):
+    if isinstance(password, six.binary_type):
+        return
+    raise TypeError("except bytes for password, not " + type(password).__name__)
```
dargueta: Grammar nit: "Accept" not "Except"

lurch: Or maybe it was a typo of "expect bytes..." ? :man_shrugging:

tzing: yeah it's typo 😆 i'd like to type 'expect'

dargueta: > Or maybe it was a typo of "expect bytes..." ?

That never occurred to me hahaha

alexpovel: So it should ultimately say 'Expected', I guess.

dargueta: That would make more sense than my suggestion.

tzing:
Still waiting for #362 merge before I continue on this branch...

PR #362, opened 2019-12-06
add options for readbytes and readtext

tzing:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

Pre discuss in #361, this PR add options for `readbytes` and `readtext` in each class.

Also it changes a behavior in `FS.open`, I found this when I was investigate if there's any other thing to be changed.
It seems no extra parameters are required for `iotools.make_stream`, and the description for `options` is *arguments for the filesystem*. So I guess it was misplaced and move options for `openbin` inside the function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.4.12] - (Unreleased)
 
+### Added
+
+- Added options for readbytes and readtext
+
 ### Changed
 
 - Start testing on PyPy. Due to [#342](https://github.com/PyFilesystem/pyfilesystem2/issues/342)
--- a/fs/base.py
+++ b/fs/base.py
@@ -221,7 +221,7 @@ class FS(object):
         path,  # type: Text
         mode="r",  # type: Text
         buffering=-1,  # type: int
-        **options  # type: Any
+        **options,  # type: Any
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.4.12] - (Unreleased)
 
+### Added
+
+- Added options for readbytes and readtext
```
lurch: "Added options parameter to readbytes and readtext functions" ?

For symmetry, maybe it's worth modifying writebytes and writetext too?

chfw:
Just an observation, readtext is as readable as read_text. But I guess it is down to historical design decision.

lurch:
And maybe it's worth adding docstrings like [this](https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/base.py#L1151) too?

lurch:
Question for @willmcgugan - looking at https://docs.pyfilesystem.org/en/latest/reference/base.html , should _all_ functions that take `path` parameter(s) also take an `options` kwargs?

@tzing Sorry for potentially opening a can of worms... :confused:  Your PR looks great so far IMHO :+1:

tzing:
ping @willmcgugan

Issue #380, opened 2020-04-08
missing readinto method for MemoryFS files?


nwh:
First, many thanks for maintaining this great library!

It appears that `MemoryFS` file objects are missing the `readinto` method.  This is tested on Python 3.6:

```python
"""Investigate the readinto method issue with MemoryFS"""

from fs import open_fs

def try_temp():
    temp_fs = open_fs("temp://")
    temp_fs.writebytes("file", b"here are some bytes")

    buff = bytearray(30)

    with temp_fs.openbin("file", mode="r") as file:
        file.readinto(buff)

    print(buff)

def try_mem_1():
    mem_fs = open_fs("mem://")
    mem_fs.writebytes("file", b"here are some bytes")

    buff = bytearray(30)

    with mem_fs.openbin("file", mode="r") as file:
        print(type(file))
        file.readinto(buff)

    print(buff)

def try_mem_2():
    mem_fs = open_fs("mem://")
    mem_fs.writebytes("file", b"here are some bytes")

    buff = bytearray(30)

    with mem_fs.open("file", mode="rb") as file:
        print(type(file))
        file.readinto(buff)

    print(buff)


try_temp() # ok, prints buff
try_mem_1()  # raises NotImplementedError
try_mem_2()  # raises NotImplementedError
```

In the case of `try_mem_2`, `file` is an instance of `fs.iotools.RawWrapper`, which appears to check if the underlying file has `readinto` by checking for `AttributeError`.  It appears that in the case of `_MemoryFile` it ends falling back to the "stub" `readinto` method, which exists, but raises `NotImplementedError`.

I am happy to prepare a PR adding `readinto` to `_MemoryFile` or checking for `NotImplementedError` in `RawWrapper`.

Thoughts?

willmcgugan:
Thanks for investigating. Implementing a readinto for memory file seems the way to go. A PR would be appreciated.

PR #381, opened 2020-04-09
Add readinto method for MemoryFS and FTPFS file objects

nwh:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Added `readinto` method for `MemoryFS` file objects.  See #380 .

### DIFF
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -239,7 +239,7 @@ class FTPFile(io.RawIOBase):
         # type: (bytearray) -> int
         data = self.read(len(buffer))
         bytes_read = len(data)
-        buffer[: len(data)] = data
+        buffer[:bytes_read] = data
         return bytes_read
 
     def readline(self, size=-1):
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -235,6 +235,13 @@ class FTPFile(io.RawIOBase):
                 remaining -= len(chunk)
         return b"".join(chunks)
 
+    def readinto(self, buffer):
+        # type: (bytearray) -> int
+        data = self.read(len(buffer))
+        bytes_read = len(data)
+        buffer[: len(data)] = data
+        return bytes_read
+
     def readline(self, size=-1):
         # type: (int) -> bytes
         return next(line_iterator(self, size))  # type: ignore
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -120,6 +120,8 @@ class _MemoryFile(io.RawIOBase):
 
     def readline(self, size=-1):
         # type: (int) -> bytes
+        if not self._mode.reading:
+            raise IOError("File not open for reading")
         with self._seek_lock():
             self.on_access()
             return self._bytes_io.readline(size)
@@ -153,6 +155,8 @@ class _MemoryFile(io.RawIOBase):
 
     def readlines(self, hint=-1):
         # type: (int) -> List[bytes]
+        if not self._mode.reading:
+            raise IOError("File not open for reading")
         with self._seek_lock():
             self.on_access()
             return self._bytes_io.readlines(hint)
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -142,6 +142,14 @@ class _MemoryFile(io.RawIOBase):
         # type: () -> bool
         return self._mode.reading
 
+    def readinto(self, b):
+        # type (bytes) -> Optional[int]
+        if not self._mode.reading:
+            raise IOError("File not open for reading")
+        with self._seek_lock():
+            self.on_access()
+            return self._bytes_io.readinto(b)
+
     def readlines(self, hint=-1):
         # type: (int) -> List[bytes]
         with self._seek_lock():
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -146,7 +146,7 @@ class _MemoryFile(io.RawIOBase):
         return self._mode.reading
 
     def readinto(self, b):
-        # type (bytes) -> Optional[int]
+        # type (bytearray) -> Optional[int]
         if not self._mode.reading:
             raise IOError("File not open for reading")
         with self._seek_lock():
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -113,6 +113,7 @@ class _MemoryFile(io.RawIOBase):
     def next(self):
         # type: () -> bytes
         with self._seek_lock():
+            self.on_access()
             return next(self._bytes_io)
 
     __next__ = next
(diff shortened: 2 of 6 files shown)


nwh:
@willmcgugan I can add a test for this to `tests_memoryfs.py` if desired.

nwh:
I tested manually via `ipython`:

```
In [2]: import fs                                                                                                       

In [3]: memfs = fs.open_fs("mem://")                                                                                    

In [4]: memfs.writebytes("file", b"here are some bytes")                                                                

In [5]: file = memfs.openbin("file", mode="rb")                                                                         

In [6]: b = bytearray(100)                                                                                              

In [7]: file.readinto(b)                                                                                                
Out[7]: 19

In [8]: b                                                                                                               
Out[8]: bytearray(b'here are some bytes\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
```

lurch:
Not directly related to this PR, but should the `next` and `readlines` methods also be calling `self.on_access()` @willmcgugan ?

willmcgugan:
@lurch Yeah, I think so!

willmcgugan:
@nwh A test would be great. The test should ideally be in `test.py` so it will also test for `readinto` in other filesystems.

fs/memoryfs.py
```diff
@@ -137,6 +137,14 @@ def readable(self):
         # type: () -> bool
         return self._mode.reading
 
+    def readinto(self, b):
+        # type (bytes) -> Optional[int]
```
willmcgugan, line 141: I think the typing should be `(bytearray) -> int`.

Although, the docs do say this [method](https://docs.python.org/3/library/io.html#io.RawIOBase.readinto) accepts a "bytes-like" object, but I don't know how to type that.

willmcgugan:
@nwh Would you mind adding those calls to `on_access` that @lurch pointed out?

fs/memoryfs.py
nwh: I chose `bytes` for this because of this note in the `typing` docs: https://docs.python.org/3/library/typing.html#typing.ByteString

> class typing.ByteString(Sequence[int])
>    A generic version of collections.abc.ByteString.
>
>    This type represents the types bytes, bytearray, and memoryview.
>
>    As a shorthand for this type, bytes can be used to annotate arguments of any of the types mentioned above.

I think a `memoryview` is also acceptable input to `readinto`.

An alternative to `bytes` here should be `typing.ByteString` (I think :)).

nwh:
@willmcgugan  Also, `_MemoryFile.read()` checks if the file is open for reading.  I could add that check to other read-like methods if you'd like.

willmcgugan:
@nwh Good idea, thanks. Don't forget to add yourself to contributors.

fs/memoryfs.py
willmcgugan: It can't be bytes, because readinto needs a mutable object.

The gold standard is probably typeshed (https://github.com/python/typeshed/blob/master/stdlib/3/io.pyi#L55) which uses bytearray.

nwh:
@willmcgugan looks like `FTPFile` also could use a `readinto` method.  I could write one.  My first thought is that `FTPFile.readinto` should simply use `FTPFile.read` and then put the bytes into the `bytearray`.  How does this sound?

nwh:
Also, I could change variable name `b` to `buff` or `buffer` for `bytearray` objects to satisfy the variable naming guidance.

willmcgugan:
> @willmcgugan looks like FTPFile also could use a readinto method. I could write one. My first thought is that FTPFile.readinto should simply use FTPFile.read and then put the bytes into the bytearray. How does this sound?

Great. Seems like a reasonable approach. Sorry this is turning in to a bigger job that anticipated.

fs/ftpfs.py
```diff
@@ -235,6 +235,13 @@ def read(self, size=-1):
                 remaining -= len(chunk)
         return b"".join(chunks)
 
+    def readinto(self, buffer):
```
lurch, line 238: @willmcgugan Should this be in the base-class, rather than ftpfs?

fs/ftpfs.py
```diff
@@ -235,6 +235,13 @@ def read(self, size=-1):
...
+        # type: (bytearray) -> int
+        data = self.read(len(buffer))
+        bytes_read = len(data)
+        buffer[: len(data)] = data
```
lurch, line 242: Any reason this isn't `buffer[:bytes_read] = data` ?

fs/ftpfs.py
nwh: `FTPFile` inherits from `io.RawIOBase` which raises `NotImplementedError` when `readinto` is called.

`fs.iotools.RawWrapper` does have a `readinto` method which provides a default implementation. However it does not catch `NotImplementedError` which is raised when a subclass of `io.RawIOBase` falls back to the `RawIOBase` default method.

It might be good for `fs.iotools.RawWrapper` to catch `NotImplementedError` for `readinto`.

fs/ftpfs.py
nwh: Ahh, good point.  I just copied the implementation from `fs.iotools.RawWrapper.readinto`.  The implementation for `fs.iotools.RawWrapper.readinto1` is also the same.  I could update all three locations if requested. Cheers!

fs/ftpfs.py
nwh: But even adding the catch in `RawWrapper` doesn't cover all cases.  `FTPFS.openbin` returns a `FTPFile` (as far as I can tell) so it is still nice to have the `readinto` method on `FTPFile`.

lurch: Fair enough, you seem to have looked into it in much more detail than me :eyes: :+1:

fs/ftpfs.py
althonos: @nwh : could you please update this (as well as the code in `RawWrapper.readinto`)? I'll merge the PR right after. Thanks!

nwh: @althonos done -- however, there is conflict in the change log.  How would you like that handled?

althonos (approved):

althonos:
@nwh : in order to fix the conflics, best would be if you could rebase against the current master branch, then update this PR by force-pushing.

nwh:
> @nwh : in order to fix the conflics, best would be if you could rebase against the current master branch, then update this PR by force-pushing.

@althonos thanks, done

althonos:
Thanks!

Issue #398, opened 2020-05-17
splitext removes a forward slash from url


simonm3:
Why does fs.path.splitext(url) remove one of the forward slashes?
```
url = 's3://simonm3/_testdata/file1.pkl'
fs.path.splitext(url), os.path.splitext(url)
```
(('s3:/simonm3/_testdata/file1', '.pkl'),
 ('s3://simonm3/_testdata/file1', '.pkl'))

Also what is the purpose of the second forward slash? The standard format of a url is that // denotes the start what urllib.urlparse calls "netloc" and includes username etc..  The pyfilesystem url instead uses // as part of the filesystem. Why not have "s3:/somefolder/somefile" rather than "s3://somefolder/somefile" ?

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax

willmcgugan:
fs.path functions aren’t intended for URLs, they operate on FS paths. https://docs.pyfilesystem.org/en/latest/concepts.html

simonm3:
IMO this is a somewhat artifiicial distinction. I am using urls as a path
as it seems convenient. Anyway seems a bit odd that splitext would do
anything other than split on ".". Removing forward slashes is not an
obvious thing it should be doing.

willmcgugan:
Double slashes don't have any meaning in a FS path. In other words "//foo/bar" references the same resource as "/foo/bar". FS path functions normalize the result so that that paths referencing the same resource will always compare equal.

They may look similar in many ways, but FS paths aren't URLs. Conflating the two will likely result in subtle breakage, you're better off sticking to the stdlib for parsing URLs.

lurch:
> IMO this is a somewhat artifiicial distinction. I am using urls as a path as it seems convenient.

That reminds me of the "I am using bytes as a string as it seems convenient" approach taken by Python2 :laughing: 

I agree with @willmcgugan that you need to keep these three things conceptually distinct: URLs, system paths, pyfilesystem fspaths. Anything else will just lead to a world of pain and confusion later on.
If you're not careful, you might end up parsing `C:/Program Files/Python/` as a URL! :stuck_out_tongue_winking_eye:

simonm3:
Well there is a specification for urls and file paths are included. This is a valid url:

`urlparse("file:c:/simonm3/_testdata")`
`ParseResult(scheme='file', netloc='', path='c:/simonm3/_testdata', params='', query='', fragment='')`

[...]

PR #400, opened 2020-05-18
document exceptional cases where the norm does not apply. #398

chfw:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Documentation / docstrings
- [ ] Tests
- [ ] Other

## Checklist

## Description

Please describe your changes here. If this fixes a bug, please link to the issue, if possible.

related to #398

### DIFF
--- a/fs/path.py
+++ b/fs/path.py
@@ -320,6 +320,9 @@ def splitext(path):
     # type: (Text) -> Tuple[Text, Text]
     """Split the extension from the path.
 
+    Please be aware that double forward slashes in your URL
+    will become single slash, i.e. s3://path -> s3:/path.
+
     Arguments:
         path (str): A path to split.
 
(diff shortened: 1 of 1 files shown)


lurch:
Presumably this "issue" affects multiple methods, and not _only_ `splitext` ?

chfw:
I would call scope creep, : P

Here is what I can find:

```
$ grep -R 'splitex' fs/
fs//path.py:    "splitext",
fs//path.py:def splitext(path):
fs//path.py:        >>> splitext('baz.txt')
fs//path.py:        >>> splitext('foo/bar/baz.txt')
fs//path.py:        >>> splitext('foo/bar/.foo')
```

willmcgugan:
@chfw The example of an `s3:` path is not a good one, because it implies you would call it with a URL. 

My gut feeling is that most of the confusion could be resolved by strategically adding the word "normalized" to the docstrings. Something like "A tuple containing the **normalized** path and the extension."

Putting a disclaimer on `splitext` probably isn't the way to go because I'm pretty sure all the methods in "fs.path" will return normalized paths, as @lurch pointed out.

lurch:
> I'm pretty sure all the methods in "fs.path" will return normalized paths

Not always - a cursory glance finds that `abspath`, `relpath` and `split` don't normalize their paths - should they? (there may be others, I didn't look at the whole file)

althonos:
Closing, since the proposing change shows a FS URL, which is *not* what the `fs.path` module is designed to work with. To discuss path normalisation for `fs.path` functions, feel free to open another issue!

PR #414, opened 2020-07-30
None docstrings fix

jcharlong:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fixes an issue when docstrings are set to `None`. Issue #415.

### DIFF
--- a/fs/_version.py
+++ b/fs/_version.py
@@ -1,3 +1,3 @@
 """Version, used in module and setup.py.
 """
-__version__ = "2.4.11"
+__version__ = "2.4.12"
--- a/fs/_version.py
+++ b/fs/_version.py
@@ -1,3 +1,3 @@
 """Version, used in module and setup.py.
 """
-__version__ = "2.4.12"
+__version__ = "2.4.11"
--- a/fs/base.py
+++ b/fs/base.py
@@ -84,7 +84,7 @@ def _new_name(method, old_name):
 """.format(
         method.__name__
     )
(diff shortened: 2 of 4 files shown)


jcharlong:
Fixes issue #415 .

fs/base.py
```diff
@@ -84,7 +84,7 @@ def _method(*args, **kwargs):
 """.format(
         method.__name__
     )
-    if hasattr(_method, "__doc__"):
+    if hasattr(_method, "__doc__") and _method.__doc__ is not None:
```
lurch, line 87: Is this the same as `if getattr(_method, "__doc__", None) is not None:` ?

fs/_version.py
```diff
@@ -1,3 +1,3 @@
 """Version, used in module and setup.py.
 """
-__version__ = "2.4.11"
+__version__ = "2.4.12"
```
althonos, line 3: Can you remove the bump for the version number ? We only update it in the commit where we publish a new release.

althonos (changes requested):
Good job, thanks! Just waiting for you to revert changes on the `fs._version` file and then I'll merge. Can you add yourself to the `CONTRIBUTORS.md` file as well if possible?

fs/_version.py
jcharlong: Fixed.

jcharlong:
Reverted the version bump & added myself to the contributors. :+1:

althonos (approved):

PR #418, opened 2020-09-01
Initial symlink creation support

cdleonard:
Implemented only in OSFS but with tests in FSTestCases

Signed-off-by: Leonard Crestez <cdleonard@gmail.com>

Not tested on windows

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -310,6 +310,21 @@ class FS(object):
 
         """
 
+    def symlink(self, src, dst):
+        # type: (Text, Text) -> None
+        """Create a symbolic link pointing to src named dst.
+
+        Arguments:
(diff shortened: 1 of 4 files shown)


patrickwolf:
Supporting symlinks seems like a very useful addition. I'm wondering if there is a deeper reason this wasn't merged in? Thanks

cdleonard:
I remember that this didn't work on windows and never bothered fixing it.

This was at a time where I was exploring if I should use this library - the answered ended up being no.

Somebody else can try to revive this and check on windows.

patrickwolf:
> This was at a time where I was exploring if I should use this library - the answered ended up being no.

Can I ask which library you ended up using @cdleonard ?

Issue #425, opened 2020-09-19
Default implementation of `FS.islink` is probably wrong


althonos:
The current implementation of `fs.base.FS.islink` is the following:

```python
    def islink(self, path):
        self.getinfo(path)
        return False
```

Because of this, and contrary to `FS.isdir` and `FS.isfile`, a filesystem implementation simply relying on the `type` attribute in the `details` namespace of `getinfo` to convey a file type will not work. This probably affects `TarFS` in this library.

PR #426, opened 2020-09-19
Implement proper symlink support in `ReadTarFS`

### LINKED ISSUE #409: Copying metadata between FS objects
Hi!

I'm looking for a way to copy/mirror data from `ReadTarFS` to `OSFS`. Everything works fine unless the archive contains symlinks. All copy/mirror methods are implemented via `copy_file_internal` function which in turn calls `openbin` method of appropriate FS-inherited object. And in `ReadTarFS` implementation of `openbin` there is ```
        if not member.isfile():
            raise errors.FileExpected(path)
```.

I get why this check is there - the library strives to be a lowest common denominator for different filesystems and symlinks are not universally-supported. I believe that my current issue can be solved by patching `if not member.isfile()` to `if not (member.isfile() or member.issym())` and using `fs.copy.copy_fs` with `on_copy` callback to alter metadata with `setinfo` call (as symlinks are simply text files with magic attribute set on them).

But it feels that a more general problem lurks there. Currently there is no sane way to customize copying (of both data and metadata) for cases when both filesystems actually support the same set of features and thus it's safe to copy some exotic types of files (like symlinks/sockets/devices/... in tar and ext4).

I like the idea behind this library but lack of such feature makes it unusable for my case. Off the top of my head I can think of [one solution](https://eli.thegreenplace.net/2016/a-polyglots-guide-to-multiple-dispatch-part-2). What do you think about it?

PS: Thanks for your time and for an interesting piece of software! =)

althonos:
## Type of changes

- [x] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I accept that @willmcgugan may be pedantic in the code review.

## Description

The current version of `ReadTarFS` does not support symlinks properly where Tar files do. This PR changes the following:

### `OSFS` like behaviour for `Info`

- [x] `Info.is_dir` is `True` for symlinks that map to existing directories.
- [x] `ReadTarFS.isdir(path)` returns True if `path` is a symlink that maps to an existing directory.
- [x] `ReadTarFS.isfile(path)` returns True if the `path` is a symlink that maps to an existing file.
- [x] `ReadTarFs.islink(path)` return True if the `path` is a symlink (that can also be dangling).

### Transparent handling of symlinks

- [x] `ReadTarFS.openbin(path)` also works if `path` is a symlink that maps to an existing file.
- [x] `ReadTarFS.listdir(path)` also works if `path` is a symlink that maps to an existing file.

##

Closes #409, #425.

### DIFF
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -348,6 +348,10 @@ class ReadTarFS(FS):
                 "is_dir": member.isdir(),
             }
 
+            if "link" in namespaces:
+                raw_info["link"] = {
+                    "target": self._decode(member.linkname) if member.issym() else None
+                }
             if "details" in namespaces:
                 raw_info["details"] = {
                     "size": member.size,
--- a/fs/tarfs.py
+++ b/fs/tarfs.py
@@ -431,11 +431,13 @@ class ReadTarFS(FS):
         except KeyError:
             six.raise_from(errors.ResourceNotFound(path), None)
 
-        if not member.isfile():
+        # TarFile.extractfile returns None if the entry is
+        # neither a file nor a symlink
+        reader =  self._tar.extractfile(member)
+        if reader is None:
             raise errors.FileExpected(path)
 
-        rw = RawWrapper(cast(IO, self._tar.extractfile(member)))
-
+        rw = RawWrapper(reader)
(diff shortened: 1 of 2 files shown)


althonos:
Moving this to milestone `v2.5.0` because this is a fundamental change.

PR #429, opened 2020-09-21
Check path for `hassyspath` and use `shutil.rmdir` if True

geoffjukes:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ x] Other

## Checklist

- [x] I accept that @willmcgugan may be pedantic in the code review :)

## Description

Minor enhancement to `removetree` when the path has a syspath and so can be removed via `shutil.rmtree`

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -12,6 +12,7 @@ import abc
 import hashlib
 import itertools
 import os
+import shutil
 import threading
 import time
 import typing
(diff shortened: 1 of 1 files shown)


fs/base.py
```diff
@@ -1219,17 +1220,20 @@ def removetree(self, dir_path):
...
+            else:
+                _dir_path = abspath(normpath(dir_path))
+                walker = walk.Walker(search="depth")
+                gen_info = walker.info(self, _dir_path)
+                for _path, info in gen_info:
+                    if info.is_dir:
+                        self.removedir(_path)
+                    else:
+                        self.remove(_path)
+                if _dir_path != "/":
```
althonos, line 1235: I think you also need to check that `_dir_path` is not `"/"` with the `shutil.rmtree` branch too, otherwise you may end up deleting the root of the filesystem and potentially preventing new files/directories from being created in that filesystem afterwards.

fs/base.py
```diff
@@ -1219,17 +1220,20 @@ def removetree(self, dir_path):
...
-            gen_info = walker.info(self, _dir_path)
-            for _path, info in gen_info:
-                if info.is_dir:
-                    self.removedir(_path)
-                else:
-                    self.remove(_path)
-            if _dir_path != "/":
-                self.removedir(dir_path)
+            if self.hassyspath(path):
+                shutil.rmtree(self.getsyspath(path))
```
althonos, line 1225: `shutil.rmtree` can potentially raise `OSError`, we would want to wrap it using the `onerror` callback, probably turning it into an `fs.errors.OperationFailed`.

althonos (changes requested):

althonos:
I think this is a good optimization, but we just need a few tests to make sure it is consistent with the current `FS` behaviour.


By the way, [`WrapReadOnly`](https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/wrap.py#L147) should be updated so that it makes the `removetree` method readonly, since this implementation does not use `remove` or `removedir` and could result in files being deleted even on a wrapped filesystem.

PR #431, opened 2020-09-26
Fix symlink handling in OSFS

### LINKED ISSUE #411: OSFS Bug: incorrect handling of symbolic links with non-existing target
OSFS won't acknowledge the existence of symlinks if the symlink points to a non-existing destination:

```python
>>> from fs import tempfs
>>> import os
>>> fs = tempfs.TempFS()                # TempFS is a OSFS subclass
>>> src, dst = "foo", "bar"
>>> os.symlink(src, fs.getsyspath(dst)) # dst is the link, pointing to src
>>> fs.touch(src)                       # ensure that src exists
>>> fs.exists(dst)                      # fs acknowledges that dst exists
True
>>> fs.islink(dst)                      # and that dst is a link
True
>>> fs.remove(src)                      # but if we remove the src file
>>> os.path.islink(fs.getsyspath(dst))  # it is still a link
True
>>> fs.exists(dst)                      # but fs does not want to acknowledge it
False
>>> fs.islink(dst)                      # this raises, rather than return False
  File "<stdin>", line 1, in <module>
  File "/.../site-packages/fs/osfs.py", line 612, in islink
    raise errors.ResourceNotFound(path)
fs.errors.ResourceNotFound: resource 'bar' not found
```

Note the exception thrown by `fs.islink()`; this goes counter to the behaviour of `fs.isfile()` and `fs.isdir()`, which both return `False` for this case.

First of all, I don't think `fs.islink()` should not throw exceptions, it should return `False` if the path truly doesn't exist.

Now, I know that symlinks are a complex subject with miriad edge cases. Symlinks can point to other symlinks, or a _parent directory_ of a path could be a symlink (e.g. `realpath` has `-P / --physical`, `-L / -- logical` modes that can lead to vastly different end results), etc. It could be that the intent is that only links that point to paths _within the root of the filesystem_ are supported.

But the OSFS [`islink()` documentation](https://docs.pyfilesystem.org/en/latest/reference/osfs.html#fs.osfs.OSFS.islink) makes no mention of such restrictions, it simply claims to return True if the given path is a symlink on disk.

### LINKED ISSUE #425: Default implementation of `FS.islink` is probably wrong

althonos:
## Type of changes

- Bug fix

## Checklist

### Closes #425

Changes the behaviour of `FS.islink` so that is never raises `ResourceNotFound`, to be consistent with `FS.isfile` and `FS.isdir`, and instead uses `FS.gettype` to check if the resource type is `symlink`.

### Closes #411

Fixes the behaviour of the following `OSFS` methods so that they work as expected:

- [x] `OSFS.islink(path)`: returns `True` on existing path to a symlink, `False` otherwise`
- [x] `OSFS.gettype(path)`: returns `ResourceType.symlink` is the path maps to a symlink (not the case before)
- [ ] `OSFS.getinfo(path)`: works even if `path` is a dangling symlink (**TODO**)
- [ ] `OSFS.scandir(dir)`: works even if `dir` contains a dangling symlink

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -981,8 +981,12 @@ class FS(object):
             bool: `True` if ``path`` maps to a symlink.
 
         """
-        self.getinfo(path)
-        return False
+        try:
+            self.getinfo(path, namespaces=["link"]).target is not None
+        except errors.MissingInfoNamespace:
+            return False  # filesystem does not support symlinks
+        except errors.ResourceNotFound:
+            return False  # path does not exist, so it can't be a link
 
     def lock(self):
         # type: () -> RLock
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -599,7 +599,7 @@ class OSFS(FS):
         self.check()
         sys_path = self._to_sys_path(path)
         with convert_os_errors("gettype", path):
-            stat = os.stat(sys_path)
+            stat = os.lstat(sys_path)
         resource_type = self._get_type_from_stat(stat)
         return resource_type
 
--- a/fs/osfs.py
+++ b/fs/osfs.py
@@ -608,8 +608,6 @@ class OSFS(FS):
         self.check()
         _path = self.validatepath(path)
         sys_path = self._to_sys_path(_path)
-        if not self.exists(path):
-            raise errors.ResourceNotFound(path)
         with convert_os_errors("islink", path):
             return os.path.islink(sys_path)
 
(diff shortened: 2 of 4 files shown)


althonos:
ping @mjpieters, @dargueta

PR #433, opened 2020-10-11
Added callback kwarg to the download() PyFilesystems method.

jw:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Added callback kwarg to the download() PyFilesystems method.

### DIFF
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -48,7 +48,8 @@ RE_WINDOWSNT = re.compile(
     (?P<name>.*)
     $
     """,
-    re.VERBOSE)
+    re.VERBOSE,
+)
 
 
 def get_decoders():
--- a/fs/appfs.py
+++ b/fs/appfs.py
@@ -30,8 +30,7 @@ __all__ = [
 
 
 class _AppFS(OSFS):
-    """Abstract base class for an app FS.
-    """
+    """Abstract base class for an app FS."""
 
     # FIXME(@althonos): replace by ClassVar[Text] once
     # https://github.com/python/mypy/pull/4718 is accepted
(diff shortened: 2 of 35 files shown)


willmcgugan:
There's no guarantee that all filesystems could support a callback in download. Unfortunately its not an option for some to support the callback and others not to, as it would break the promise that all filesystem implementations work the same way.

If you need this functionality, you can still copy using open files, although you may miss out on any optimizations from filesystems with a custom download method.

I noticed you probably formatted the code with black. I use it everywhere myself, and it might be nice for a single PR just to blackify the entire codebase.

jw:
Thanks Will, but even when the callback is used in a kwargs?  The implementations that do not support the feature could just silently (maybe with a warning) discard it?

If you think this is not the correct approach, I would understand.

Hopefully I can use the `fs.copy` approach you propose.  I guess the only way to have some callback like behaviour is via the `on_copy` keyword?  And this is only called after a complete copy of a file.  So large files would show no progress to the user?  Am I missing something?

willmcgugan:
> Thanks Will, but even when the callback is used in a kwargs? The implementations that do not support the feature could just silently (maybe with a warning) discard it?

Granted it wouldn't break at a signature level, but it would mean that for some filesystems you wouldn't get that callback and the code may not work as intended. Bear in mind there are many more PyFS filesystems out there that aren't in the core package. 

I'd suggest implementing your own copy routine that opens two files, one for reading, and one for writing. Then you have complete visibility of the bytes being copied.

jw:
Ok, I'll go for your file copy routine.  Thx!

PR #435, opened 2020-10-13
Require `FS.openbin` to return a file-handle with in binary mode

### LINKED ISSUE #406: ftpfs possible issue with openbin
Recently, while trying out the ftpfs, I've noticed that to open a binary file-like object for writing by some third-party packages I have to do this:

    with open_fs("ftp://anonymous:anonymous@localhost:8021") as fs:
        with fs.openbin("test.bin", "wb") as f:
            ...

Note in the above the mode param used in `openbin` includes `b`: Looking at the docs for the mode param it suggests the `b` is implied via `openbin`, but `b` isn't implicitly *added* to the mode of the file-like that is returned by `openbin` - this means that, unless `b` is explicitly provided as above, third-party packages that inspect the mode of the file-like may complain that it is not opened in binary mode.

Should `openbin` not reflect the implied `b` by implicitly adding `b` to the mode?

althonos:
## Type of changes

- Bug fix
- Tests

## Checklist

## Description

- Add a test for `openbin` method in `FSTestCases` ensuring the returned object has a `mode` attribute containing the "b" flag.
- Fix `FTPFS.openbin` not implicitly converting mode to binary mode. Closes #406 .
- Add missing `mode` attribute to `_MemoryFile` objects created by `MemoryFS.openbin`.

### DIFF
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -77,6 +77,7 @@ class _MemoryFile(io.RawIOBase):
 
     @property
     def mode(self):
+        # type: () -> Text
         return self._mode.to_platform_bin()
 
     @contextlib.contextmanager
--- a/fs/memoryfs.py
+++ b/fs/memoryfs.py
@@ -75,6 +75,10 @@ class _MemoryFile(io.RawIOBase):
         _template = "<memoryfile '{path}' '{mode}'>"
         return _template.format(path=self._path, mode=self._mode)
 
+    @property
+    def mode(self):
+        return self._mode.to_platform_bin()
+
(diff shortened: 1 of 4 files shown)


fs/memoryfs.py
```diff
@@ -75,6 +75,10 @@ def __str__(self):
         _template = "<memoryfile '{path}' '{mode}'>"
         return _template.format(path=self._path, mode=self._mode)
 
+    @property
+    def mode(self):
```
willmcgugan, line 79: Could use typing...

willmcgugan (approved):
LGTM

PR #439, opened 2020-11-05
Parsing compatibility of the FTP `LIST` command for Windows servers

### LINKED ISSUE #438: Add support for different FTP LIST locales (Windows CE exclusive?)
We are using pyfilesystem2 to access a German locale FTP server on an MS Windows CE operating system. I cannot pinpoint whether it is the locale, the OS, or a different factor altogether, but the FTP LIST command, which is apparently used almost everywhere in the library, yields its data in a different format than Windows NT. Thus, `FTPFS.read_dir` fails to parse any items and always returns an empty dict and most operations fail.

To be precise, the LIST output only differs slightly in that it uses 24 hour days rather than 12 hour AM/PM days.

WinNT:
    `11-02-18  02:12PM       <DIR>          images`
    `11-02-18  03:33PM                 9276 logo.gif`

WinCE:
    `11-02-18  14:12       <DIR>          images`
    `11-02-18  15:33                 9276 logo.gif`

We would like to see support for that format added, either as an extension to the "WindowsNT" parsing, or as a new kind of response parsing. If this is an acceptable change, we'd create the suitable PR for it. It would be great if this fix would make it into 2.4.12, as we'd have to keep using a custom fix in our code until 2.4.13 otherwise. 

Thanks!

atollk:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Extended parsing compatibility of the FTP `LIST` command for Windows servers. Fixes #438 .
The parsing process now properly supports 24-hour time format, both with and without leading zeros. Unit tests were adapted accordingly.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,3 +13,4 @@ Many thanks to the following developers for contributing to this project:
 - [Will McGugan](https://github.com/willmcgugan)
 - [Zmej Serow](https://github.com/zmej-serow)
 - [Morten Engelhardt Olsen](https://github.com/xoriath)
+- [Andreas Tollkötter](https://github.com/atollk)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -2,6 +2,7 @@
 
 Many thanks to the following developers for contributing to this project:
 
+- [Andreas Tollkötter](https://github.com/atollk)
 - [C. W.](https://github.com/chfw)
 - [Diego Argueta](https://github.com/dargueta)
 - [Geoff Jukes](https://github.com/geoffjukes)
@@ -9,8 +10,7 @@ Many thanks to the following developers for contributing to this project:
 - [Justin Charlong](https://github.com/jcharlong)
 - [Louis Sautier](https://github.com/sbraz)
 - [Martin Larralde](https://github.com/althonos)
+- [Morten Engelhardt Olsen](https://github.com/xoriath)
 - [Nick Henderson](https://github.com/nwh)
 - [Will McGugan](https://github.com/willmcgugan)
 - [Zmej Serow](https://github.com/zmej-serow)
-- [Morten Engelhardt Olsen](https://github.com/xoriath)
-- [Andreas Tollkötter](https://github.com/atollk)
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -149,12 +149,8 @@ def decode_linux(line, match):
     return raw_info
 
 
-def _decode_windowsnt_time(mdate, mtime):
-    if len(mtime.split(":")[0]) == 1:
-        mtime = "0" + mtime
-    return _parse_time(
-        mdate + " " + mtime, formats=["%d-%m-%y %I:%M%p", "%d-%m-%y %H:%M"]
-    )
+def _decode_windowsnt_time(mtime):
+    return _parse_time(mtime, formats=["%d-%m-%y %I:%M%p", "%d-%m-%y %H:%M"])
 
 
 def decode_windowsnt(line, match):
@@ -183,7 +179,7 @@ def decode_windowsnt(line, match):
         raw_info["details"]["size"] = int(match.group("size"))
 
     modified = _decode_windowsnt_time(
-        match.group("modified_date"), match.group("modified_time")
+        match.group("modified_date") + " " + match.group("modified_time")
     )
     if modified is not None:
         raw_info["details"]["modified"] = modified
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -95,10 +95,16 @@ def _find_suitable_format(t, formats):
 
 
 def _parse_time(t, formats):
-    frmt = _find_suitable_format(t, formats)
-    if frmt is None:
+    _t = None
+    for frmt in formats:
+        try:
+            _t = time.strptime(t, frmt)
+            if t:
+                break
+        except ValueError:
+            continue
+    if not _t:
         return None
-    _t = time.strptime(t, frmt)
 
     year = _t.tm_year if _t.tm_year != 1900 else time.localtime().tm_year
     month = _t.tm_mon
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -85,7 +85,6 @@ def parse_line(line):
 
 
 def _parse_time(t, formats):
-    _t = None
     for frmt in formats:
         try:
             _t = time.strptime(t, frmt)
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -145,11 +145,11 @@ def decode_linux(line, match):
     return raw_info
 
 
-def _decode_windowsnt_time(date, time):
-    while len(time.split(":")[0]) < 2:
-        time = "0" + time
+def _decode_windowsnt_time(mdate, mtime):
+    if len(mtime.split(":")[0]) == 1:
+        mtime = "0" + mtime
     return _parse_time(
-        date + " " + time, formats=["%d-%m-%y %I:%M%p", "%d-%m-%y %H:%M"]
+        mdate + " " + mtime, formats=["%d-%m-%y %I:%M%p", "%d-%m-%y %H:%M"]
     )
 
 
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -41,14 +41,17 @@ RE_LINUX = re.compile(
 RE_WINDOWSNT = re.compile(
     r"""
     ^
-    (?P<modified>.*?(AM|PM))
-    \s*
-    (?P<size>(<DIR>|\d*))
-    \s*
+    (?P<modified_date>\S+)
+    \s+
+    (?P<modified_time>\S+(AM|PM)?)
+    \s+
+    (?P<size>(<DIR>|\d+))
+    \s+
     (?P<name>.*)
     $
     """,
-    re.VERBOSE)
+    re.VERBOSE,
+)
 
 
 def get_decoders():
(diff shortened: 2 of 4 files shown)


fs/_ftp_parse.py
```diff
@@ -138,12 +145,22 @@ def decode_linux(line, match):
     return raw_info
 
 
+def _decode_windowsnt_time(date, time):
```
lurch, line 148: Might be worth renaming these parameters to `mdate` and `mtime`, to avoid potential clashes with the `time` module-name?

fs/_ftp_parse.py
```diff
@@ -138,12 +145,22 @@ def decode_linux(line, match):
...
+    while len(time.split(":")[0]) < 2:
```
lurch, line 149: Seems unlikely that `len(time.split(":")[0])` would ever be 0, so I guess the `while` could be changed to an `if` ?

tests/test_ftp_parse.py
```diff
@@ -164,39 +165,50 @@ def test_decode_linux(self, mock_localtime):
...
         directory = """\
+unparsable line
 11-02-17  02:00AM       <DIR>          docs
 11-02-17  02:12PM       <DIR>          images
-11-02-17  02:12PM       <DIR>          AM to PM
+11-02-17 02:12PM <DIR> AM to PM
 11-02-17  03:33PM                 9276 logo.gif
+05-11-20   22:11  <DIR>       src
+11-02-17   01:23       1      12
+11-02-17    4:54                 0 icon.bmp
```
lurch, line 175: Just for the sake of completeness, might be worth also adding e.g.
```
11-02-17    4:54AM                 0 icon.gif
11-02-17    4:54PM                 0 icon.png
```
? :shrug:

lurch:
If you're feeling generous, perhaps https://pydoc.net/grizzled/0.9.4/grizzled.net.ftp.parse/ and https://github.com/stevemayne/pyftpparser would appreciate similar patches? :slightly_smiling_face:

fs/_ftp_parse.py
atollk: Done

fs/_ftp_parse.py
atollk: Done

tests/test_ftp_parse.py
atollk: Done

atollk:
I created an issue for pyftpparser; I'll have to see whether I'll also do the PR there.
Regarding grizzled, the Github repo seems to be out of sync with that documentation, so it seems like it would be much more work to figure out, where the code would even go.

atollk:
Hey, any update on this? The PR seems to be ready to merge.

CONTRIBUTORS.md
```diff
@@ -13,3 +13,4 @@ Many thanks to the following developers for contributing to this project:
 - [Will McGugan](https://github.com/willmcgugan)
 - [Zmej Serow](https://github.com/zmej-serow)
 - [Morten Engelhardt Olsen](https://github.com/xoriath)
+- [Andreas Tollkötter](https://github.com/atollk)
```
dargueta, line 16: The contributors list is supposed to be sorted alphabetically, so please put your name at the top since it starts with A.

lurch:
@atollk I know it's not in code that you've modified in this PR, but I think it might be worth making these tweaks to the `_parse_time` function:
```patch
--- a/fs/_ftp_parse.py
+++ b/fs/_ftp_parse.py
@@ -85,12 +85,12 @@ def parse_line(line):
 
 
 def _parse_time(t, formats):
-    t = " ".join(token.strip() for token in t.lower().split(" "))
-
     _t = None
     for frmt in formats:
         try:
             _t = time.strptime(t, frmt)
+            if _t:
+                break
         except ValueError:
             continue
     if not _t:
```

1. Due to the way that `split` and `join` work, _all_ the removed line effectively does is `t = t.lower()` - but that isn't necessary as Python's `strptime` is equally happy with mixed-case strings (and it copes with multiple spaces too, which this line wasn't removing in any case!)
2. The second change makes it more explicit that it stops looping through formats once it finds one that works (rather than implicitly assuming that there'll only be one format that won't raise a `ValueError`)

fs/_ftp_parse.py
```diff
@@ -81,17 +84,21 @@ def parse_line(line):
...
             continue
-    if not _t:
+    return None
+
+
+def _parse_time(t, formats):
+    frmt = _find_suitable_format(t, formats)
+    if frmt is None:
         return None
+    _t = time.strptime(t, frmt)
```
lurch, line 101: Hmmm, seems a bit wasteful to do `time.strptime` (at least) twice? Once in `_find_suitable_format` and again in `_parse_time`.

lurch:
@atollk I've looked into it a bit further, and it seems like `strptime` **doesn't** require the leading '0' to be added to the `%H` format:
```python
>>> import time
>>> time.strptime("11-02-17    04:54".strip(), "%d-%m-%y %H:%M")
time.struct_time(tm_year=2017, tm_mon=2, tm_mday=11, tm_hour=4, tm_min=54, tm_sec=0, tm_wday=5, tm_yday=42, tm_isdst=-1)
>>> time.strptime("11-02-17    4:54".strip(), "%d-%m-%y %H:%M")
time.struct_time(tm_year=2017, tm_mon=2, tm_mday=11, tm_hour=4, tm_min=54, tm_sec=0, tm_wday=5, tm_yday=42, tm_isdst=-1)
```
which makes your `mtime = "0" + mtime` unnecessary?

In fact, with this change:
```patch
@@ -145,12 +145,8 @@ def decode_linux(line, match):
     return raw_info
 
 
-def _decode_windowsnt_time(mdate, mtime):
-    if len(mtime.split(":")[0]) == 1:
-        mtime = "0" + mtime
-    return _parse_time(
-        mdate + " " + mtime, formats=["%d-%m-%y %I:%M%p", "%d-%m-%y %H:%M"]
-    )
+def _decode_windowsnt_time(mtime):
+    return _parse_time(mtime, formats=["%d-%m-%y %I:%M%p", "%d-%m-%y %H:%M"])
 
 
 def decode_windowsnt(line, match):
@@ -179,7 +170,7 @@ def decode_windowsnt(line, match):
         raw_info["details"]["size"] = int(match.group("size"))
 
     modified = _decode_windowsnt_time(
-        match.group("modified_date"), match.group("modified_time")
+        match.group("modified_date") + " " + match.group("modified_time")
     )
     if modified is not None:
         raw_info["details"]["modified"] = modified
```
all the unit-tests still pass, and it has the benefit of making the `_decode_windowsnt_time` function signature more similar to the `_decode_linux_time` function signature?
Feel free to ignore this comment if you feel I'm being too picky!

fs/_ftp_parse.py
atollk: That statement is correct though in my opinion, "a bit" is negligible. A call to `strptime` costs about 4 microseconds on my machine. I find `_find_suitable_format` easier to read than something like `_decode_time_in_suitable_format` and I'd argue that readability should always be the higher priority over resource efficiency in Python - otherwise one should be using C anyway.

That being said, if you feel like that has to be changed to be accepted, I'll change it.

atollk:
> and it seems like strptime doesn't require the leading '0' to be added to the %H format

That's a fair point. I just followed the documentation there without actually trying it.

fs/_ftp_parse.py
lurch: I'll leave it up to the other maintainers...

althonos: I'd be in favour of a single function as well. I don't think it hurts readability that much, and the functions are quite highly coupled anyway.

atollk: Done

atollk:
Hi. Any update on this?

fs/_ftp_parse.py
```diff
@@ -82,12 +85,12 @@ def parse_line(line):
 
 
 def _parse_time(t, formats):
-    t = " ".join(token.strip() for token in t.lower().split(" "))
-
     _t = None
     for frmt in formats:
         try:
             _t = time.strptime(t, frmt)
+            if t:
```
lurch, line 92: I suspect you meant to do `if _t:` here?

atollk: True. But I guess this just shows that the line is entirely noop since it wasn't ever reached during the tests. Are you sure you want to have it in there? Then I'll commit a fix for the typo.

lurch: As I mentioned in an earlier comment on this issue, the current code (and tests) assume that in the list of `formats`, there'll only be a single format that doesn't raise a `ValueError`. But IMHO it makes sense to (also) stop looping after the first "valid" format, rather than continuing to loop through all of them?

atollk: Sure, I understand that. My suggestion would be to just omit the `if _t` as, when that piece of code is reached, `_t` will always have a valid value.

fs/_ftp_parse.py
```diff
@@ -82,15 +85,14 @@ def parse_line(line):
...
-
     _t = None
     for frmt in formats:
         try:
             _t = time.strptime(t, frmt)
+            break
         except ValueError:
             continue
-    if not _t:
+    else:
```
lurch, line 95: Oooh, today I just learned about the `for ... else ... ` syntax in Python :slightly_smiling_face:

fs/_ftp_parse.py
```diff
@@ -82,15 +85,14 @@ def parse_line(line):
 
 
 def _parse_time(t, formats):
-    t = " ".join(token.strip() for token in t.lower().split(" "))
-
     _t = None
```
lurch, line 88: As the `if not _t` line has been removed, I guess this `_t = None` can be removed too?

atollk: You're absolutely right, sorry.

lurch:
LGTM :+1:  ping @althonos

atollk:
Happy new year :)
Any update on this? @althonos

althonos:
LGTM @atollk ! We should be able to get this in the next debug release (once we synchronize on how we want to manage releases there). Cheers!

PR #442, opened 2021-01-13
Export 'FS' in 'fs.base'

ToddG:
## Type of changes

- [X ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Please describe your changes here. If this fixes a bug, please link to the issue, if possible.

My project(s) are using static analysis tools such as pytype and mypy. MyPy in particular requires type annotations that I cannot access with the current code. Here is an example:

```python
from fs.base import FS

class Foo:
    def __init__(self, filesystem: FS):
        self._fs = filesystem
```

This commit simply adds `FS` to the exported modules in `fs/__init__.py`.

### DIFF
--- a/fs/__init__.py
+++ b/fs/__init__.py
@@ -8,5 +8,6 @@ from .enums import ResourceType, Seek
 from .opener import open_fs
 from ._fscompat import fsencode, fsdecode
 from . import path
+from . import base
 
-__all__ = ["__version__", "ResourceType", "Seek", "open_fs"]
+__all__ = ["__version__", "ResourceType", "Seek", "open_fs", "base"]


althonos:
Hi @ToddG , i can't really understand the reason for this to be changed! The `fs.base` package is a public module with a stable API, and that's were the `FS` base class resides, so why would you expect it to be exposed elsewhere? AFAIK, a static analyzer is fine with you importing it from `fs.base`.

ToddG:
@althonos - You are correct. This was a misunderstanding on my end. Thank you for both the library and the quick response.

PR #445, opened 2021-01-20
Update base.fs.upload docs to reflect exceptions thrown

eelkevdbos:
## Type of changes

- Documentation / docstrings

## Checklist

## Description

A small docs improvement I came across while using the `fs.base.upload` method. It reflects the method throwing `ResourceNotFound` if the path you are uploading to does not exist.

### DIFF
--- a/fs/test.py
+++ b/fs/test.py
@@ -1494,6 +1494,10 @@ class FSTestCases(object):
         with self.fs.open("foo", "rb") as f:
             data = f.read()
         self.assertEqual(data, b"bar")
+        
+        # upload to non-existing path (/foo/bar)
+        with self.assertRaises(errors.ResourceNotFound):
+            self.fs.upload("/foo/bar/baz", bytes_file)
 
     def test_upload_chunk_size(self):
         test_data = b"bar" * 128
--- a/fs/test.py
+++ b/fs/test.py
@@ -1495,9 +1495,9 @@ class FSTestCases(object):
             data = f.read()
         self.assertEqual(data, b"bar")
         
-        # upload to non-existing path (/foo/bar)
+        # upload to non-existing path (/spam/eggs)
         with self.assertRaises(errors.ResourceNotFound):
-            self.fs.upload("/foo/bar/baz", bytes_file)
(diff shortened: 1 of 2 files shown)


althonos:
Thanks for the PR @eelkevdbos !

Could you possibly add a test to reflect this to `fs.test`? If this is an explicitly advertised behaviour then we need to check all filesystems behave so.

fs/test.py
```diff
@@ -1494,6 +1494,10 @@ def test_upload(self):
         with self.fs.open("foo", "rb") as f:
             data = f.read()
         self.assertEqual(data, b"bar")
+        
+        # upload to non-existing path (/foo/bar)
+        with self.assertRaises(errors.ResourceNotFound):
+            self.fs.upload("/foo/bar/baz", bytes_file)
```
althonos, line 1500: Ok, I'm clearly nitpicking here, but in that test isn't `/foo` an existing file here? There is a risk that a different error is raised by some implementors (`DirectoryExpected` for instance).

Could you try with a path that truly doesn't exist (e.g. `/spam/eggs`) ?

althonos (changes requested):

fs/test.py
eelkevdbos: I somehow missed that ;-) Maybe it is even more sufficient to assert calling underlying `openbin`, since that already has tests that assert raising the `ResourceNotFound`

althonos: Actually, some filesystem may have a more efficient method that does *not* use `openbin`, so only checking `openbin` would not be a complete test.

E.g. for my [`fs.sshfs`](https://github.com/althonos/fs.sshfs) implementation, I ~~could~~ **should** implement the `SSHFS.upload` method in a way that uses [`SFTP.putfo`](http://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.putfo) to upload a file rather than using the default method `openbin`, which calls [`SFTP.open`](http://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.open)!

althonos:
Thanks a bunch!

Issue #437, opened 2020-10-21
FTPS


andrewerf:
Hi!
Does it support ftp through SSL\TLS?
I've looked though code briefly and have found you're using ftplib.FTP, so the direct answer is no? But maybe there are some plugins/extensions for ftps?

althonos:
Hi @andrewerf ,

I don't think it's supported by `FTPFS` right now, but the `ftplib` module from the standard library [seems already capable of doing that](https://stackoverflow.com/questions/5534830/ftpes-ftp-over-explicit-tls-ssl-in-python) using [`ftplib.FTP_TLS`](https://docs.python.org/3/library/ftplib.html#ftplib.FTP_TLS). 

PRs are welcome! :wink:

PR #449, opened 2021-01-31
FTPS support

odgalvin:
## Type of changes

- New feature

## Checklist

## Description

Added FTPS (FTP over TLS) support to FTPFS. Addresses #437.

* Try to import FTP_TLS from the standard library, but if it's not available just default to a plain FTP connection, and set the `tls` attribute to False
* Add a boolean `tls` argument to the constructor, and as an attribute
* Add details and examples to the class docstring
* Show the ftps:// scheme in the URL when TLS is enabled
* Enable TLS when the ftps:// scheme is used in open/open_fs
* Add basic test

### DIFF
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -473,7 +473,7 @@ class FTPFS(FS):
             _ftp.connect(self.host, self.port, self.timeout)
             _ftp.login(self.user, self.passwd, self.acct)
             try:
-                _ftp.prot_p()
+                _ftp.prot_p()  # type: ignore
             except AttributeError:
                 pass
             self._features = {}
--- a/fs/ftpfs.py
+++ b/fs/ftpfs.py
@@ -14,6 +14,11 @@ import typing
 from collections import OrderedDict
 from contextlib import contextmanager
 from ftplib import FTP
+
+try:
+    from ftplib import FTP_TLS
+except ImportError:
+    FTP_TLS = None
 from ftplib import error_perm
 from ftplib import error_temp
 from typing import cast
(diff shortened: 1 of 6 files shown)


fs/ftpfs.py
```diff
@@ -390,6 +420,7 @@ def __init__(
         self.timeout = timeout
         self.port = port
         self.proxy = proxy
+        self.tls = tls
```
althonos, line 424: PyFilesystem is quite explicit about things most of the time, so I think we should have a hard error here if `tls=True` and `FTP_TLS is None`. Could you please update the code so that it raises `fs.errors.CreateFailed("FTP over TLS not supported")` here?

fs/ftpfs.py
```diff
@@ -432,11 +463,17 @@ def _parse_features(cls, feat_response):
     def _open_ftp(self):
         # type: () -> FTP
         """Open a new ftp object."""
-        _ftp = FTP()
+        if self.tls and FTP_TLS:
+            _ftp = FTP_TLS()
+        else:
+            self.tls = False
+            _ftp = FTP()
```
althonos, line 470: Consequently, here, you could just replace these lines with:
```python
_ftp = FTP_TLS() if self.tls else FTP()
```

fs/ftpfs.py
althonos: Even better, we could chain the eventual import error into the `CreatedFailed`, by importing first with:
```python
try:
    from ftplib import FTP_TLS
except ImportError as err:
    FTP_TLS = err
```
and then in `FTPFS.__init__`:
```python
if self.tls and isinstance(FTP_TLS, Exception):
    six.raise_from(fs.errors.CreateFailed("FTP over TLS not supported"), FTP_TLS)
```

althonos (changes requested):
Many thanks for taking the time to implement this! I just have a change request for platform where FTP_TLS is not available, and we should be in the clear.

odgalvin:
Good idea! I just pushed a commit with that changed.

althonos (approved):

althonos:
Great! Now all that's left is to add a few `# type: ignore` comments to `fs/ftpfs.py` to make the linter happy, and to update the `CONTRIBUTORS.md` and `CHANGELOG.md` files :wink:

althonos:
Many thanks!

odgalvin:
No problem, thanks!

Secret158:
You are the best

PR #457, opened 2021-03-19
In fs.ftpfs.FTPFS.upload, replaced `self._manage_ftp` by `self.ftp`.

### LINKED ISSUE #455: fs.ftpfs.FTPFS.upload opens new connections
Hi,
I'm looking at this function:
https://github.com/PyFilesystem/pyfilesystem2/blob/64d7a52e1f3037f33937f91e9db812cafdbbafc0/fs/ftpfs.py#L813

`_manage_ftp` always opens a new FTP connection, together with user login, "FEAT" command and so on. This is done everytime one wants to copy a file to an FTPFS. Is there a good reason why that code doesn't just use `self._ftp`, the connection that should be open for the lifetime of the FTPFS object? Otherwise, I'd like to create a PR to fix that.

atollk:
## Type of changes

- Other

 (performance improvement)

## Checklist

## Description

Fixes #455

`_manage_ftp` opens a new FTP connection, which is not necessary in this case, as `self._lock` is obtained anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Make `FTPFile`, `MemoryFile` and `RawWrapper` accept [`array.array`](https://docs.python.org/3/library/array.html)
   arguments for the `write` and `writelines` methods, as expected by their base class [`io.RawIOBase`](https://docs.python.org/3/library/io.html#io.RawIOBase).
 - Various documentation issues, including `MemoryFS` docstring not rendering properly.
+- Avoid creating a new connection on every call of `FTPFS.upload`. Closes [#455](https://github.com/PyFilesystem/pyfilesystem2/issues/455).
 
 
 ## [2.4.12] - 2021-01-14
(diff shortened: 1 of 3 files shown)


atollk:
I have left CHANGELOG.md unchanged for now until the full scope of this PR has been set in stone.

I have not added tests since I wouldn't really know how, except for maybe hijacking `ftplib` itself to check that no inapproprate FTP commands are sent. If that's desired, let me know and I can add it.

althonos:
I think the tests here are already present, as long as the function is tested i think we are fine. Maybe mocking the `_manage_ftp` with [`assert_not_called`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called).

The question here is whether to extend this to other methods like `download`, `readbytes`, `writebytes`, etc. For instance, in `fs.sshfs` I implemented all `SSHFS` methods with a lock, and only open a new connection when a file is obtained with `open` or `openbin`. This is likely faster when the `FS` object is used sequentially, but I'm not sure about parallel performance.

I think we will need benchmarking to know the right approach.

atollk:
Sure thing. Is there already an established framework for benchmarks in PyFs or some form of CI?

althonos:
I don't think Python has an established benchmarking framework, unfortunately. 

In order to avoid blowing the scope of this PR, I'll just merge these changes, and we can consider a performance improvement of larger scale after we made sure which strategy is the most efficient.

PR #460, opened 2021-03-20
fs.copy.copy_*_if

atollk:
## Type of changes

- New feature
- Tests

## Checklist

## Description

Added _copy_if functions to fs.copy module.

The free functions copy_file, copy_file_if_newer, copy_dir, copy_dir_if_newer, copy_fs, and copy_fs_if_newer now are all implemented in terms of copy_file_if, copy_dir_if, and copy_fs_if.
Unit Tests for this change will be created in a future commit.

This change streamlined the code by removing duplication in logic and clustering similar logic closer together.

Also, though this has not been tested, copy_dir_if_newer should be faster in this implementation when copying to a dst_dir that contains a lot of files that are not present in the respective src_dir.

Finally, a bug was fixed that was caused by a false lookup in the dictionary of file infos from dst_fs. This fix could cause unnecessary copies and therefore a decrease in performance.

### DIFF
--- a/fs/copy.py
+++ b/fs/copy.py
@@ -5,7 +5,7 @@ from __future__ import print_function, unicode_literals
 
 import typing
 
-from .errors import FSError
+from .errors import ResourceNotFound
 from .opener import manage_fs
 from .path import abspath, combine, frombase, normpath
 from .tools import is_thread_safe
@@ -41,9 +41,7 @@ def copy_fs(
             a single-threaded copy.
 
     """
-    return copy_dir(
-        src_fs, "/", dst_fs, "/", walker=walker, on_copy=on_copy, workers=workers
-    )
+    return copy_fs_if(src_fs, dst_fs, "always", walker, on_copy, workers)
 
 
 def copy_fs_if_newer(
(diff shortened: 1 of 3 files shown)


atollk:
Imho, `copy_*_if_newer` should be marked as deprecated if this change is merged.

althonos:
Nice work! I'll wait for your unit tests to be added.
 
I'm thinking we should soft-deprecate the `copy_if_newer` functions eventually, since they are just aliasing now, and it's confusing for the end user that we provide them for some conditions (`newer`) but not for the remaining ones.

fs/copy.py
```diff
@@ -332,54 +363,128 @@ def copy_dir_if_newer(
...
+            )
+
+    elif condition == "exists":
+        return dst_fs.exists(dst_path)
+
+    elif condition == "not_exists":
+        return not dst_fs.exists(dst_path)
+
+    else:
+        raise ValueError(condition + "is not a valid copy condition.")
```
althonos, line 490: Please use string formatting here, in case `condition` is not a string this will fail (plus, you are missing a whitespace :wink: )

fs/copy.py
```diff
@@ -74,38 +72,61 @@ def copy_fs_if_newer(
...
+
+    If ``condition`` has the value ``"older"``, the last modification time
+    of the source file must be older than that of the destination file.
+    If either file has no modification time, the copy is performed always.
+
+    If ``condition`` has the value ``"exists"``, the source file is only
+    copied if a file of the same path already exists in ``dst_fs``.
+
+    If ``condition`` has the value ``"not_exists"``, the source file is only
+    copied if no file of the same path already exists in ``dst_fs``.
```
althonos, line 104: Could you perhaps see if this looks better as a reStructuredText definition list? I'm thinking it may render a bit more nicely in the HTML documentation. This is done in `fs.glob` for instance (https://docs.pyfilesystem.org/en/latest/globbing.html):
```rst
Here's a summary of glob patterns:

``*``
    Matches all files in the current directory.
``*.py``
    Matches all .py file in the current directory.
``*.py?``
    Matches all .py files and .pyi, .pyc etc in the currenct directory.
```
(i'm noticing a typo there btw ahah)

fs/copy.py
```diff
@@ -126,105 +147,135 @@ def copy_file(
...
+    else:
+        _copy_locked()
 
 
 def copy_structure(
     src_fs,  # type: Union[FS, Text]
     dst_fs,  # type: Union[FS, Text]
     walker=None,  # type: Optional[Walker]
+    src_root="/",  # type: Text
+    dst_root="/",  # type: Text
```
althonos, line 278: A question here: do we really need these new `src_root` and `dst_root` parameters? It's already possible to copy subdirectories using the `opendir` and `makedir` to get a `SubFS`:
```python
src_fs = fs.open_fs("...")
dst_fs = fs.open_fs("...")
fs.copy_structure(fs.opendir("foo"), fs.makedir("bar", recreate=False))
```
I don't know if we really need to add 2 extra parameters, it may get confusing if there is more than one way to do the same thing (aka *There should be one-- and preferably only one --obvious way to do it.*).

althonos (changes requested):

fs/copy.py
atollk: Sure, if you feel like those parameters should be covered by `SubFS` instead, I can do that. Although the question is whether the break with the other functions in the module isn't a more confusing design. (e.g. `copy_dir` takes both `src_fs` and `src_path`).

atollk:
@althonos Thanks. I already added unit tests, I just forgot to remove that part from the description, sorry. It looks like the lines that are not covered anymore are the `_if_newer` functions which are just one-line wrappers now.

fs/copy.py
atollk: Done!

fs/copy.py
atollk: Done. Didn't know that Python 2 already supported that.

fs/copy.py
althonos: True. Then in that case, maybe having the signature be `copy_structure(src_fs, src_path, dst_fs, dst_path, walker=None)` would be more consistent?

fs/copy.py
```diff
@@ -54,100 +53,168 @@ def copy_fs_if_newer(
     workers=0,  # type: int
 ):
     # type: (...) -> None
-    """Copy the contents of one filesystem to another, checking times.
+    """Use ``copy_fs_if`` instead."""
```
althonos, line 56: For documenting deprecation, please keep the previous summary, and add the [`deprecated`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) Sphinx directive:
```python
"""Copy the contents of one filesystem to another, checking times.

.. deprecated:: 2.5.0
   Use `~fs.copy_fs_if` with ``condition="newer"`` instead.
"""
```

fs/copy.py
```diff
@@ -54,100 +53,168 @@ def copy_fs_if_newer(
...
+    warnings.warn(
+        DeprecationWarning("copy_fs_if_newer is deprecated. Use copy_fs_if instead.")
+    )
```
althonos, line 59: I think right way to raise warnings is (according to the stdlib [docs](https://docs.python.org/3/library/warnings.html)):
```python
warnings.warn("copy_fs_if_newer id deprecated. Use copy_fs_if instead", DeprecationWarning)
```

fs/copy.py
```diff
@@ -311,75 +325,132 @@ def copy_dir_if_newer(
     workers=0,  # type: int
 ):
     # type: (...) -> None
-    """Copy a directory from one filesystem to another, checking times.
+    """Use ``copy_dir_if`` instead."""
+    warnings.warn(
+        DeprecationWarning("copy_dir_if_newer is deprecated. Use copy_dir_if instead.")
+    )
```
althonos, line 331: Same as earlier.

fs/copy.py
```diff
@@ -311,75 +325,132 @@ def copy_dir_if_newer(
...
-            a single-threaded copy.
+            condition (str): Name of the condition to check for each file.
+            walker (~fs.walk.Walker, optional): A walker object that will be
+                used to scan for files in ``src_fs``. Set this if you only want
+                to consider a sub-set of the resources in ``src_fs``.
+            on_copy (callable):A function callback called after a single file copy
+                is executed. Expected signature is ``(src_fs, src_path, dst_fs,
+                dst_path)``.
+            workers (int): Use ``worker`` threads to copy data, or ``0`` (default) for
+                a single-threaded copy.
```
althonos, line 382: Can you fix the indentation for these arguments?

althonos (changes requested):
Some small fixes related to the deprecation of older functions, and we should be good to go!

fs/copy.py
atollk: I agree but that would be a breaking change. Would that be acceptable?

atollk: @althonos Any opinion on this? I could easily change the signature of that function but it feels to me like that would break too much existing code.

althonos: Let's stick with the non-breaking signature for now, it looks like we are going to have to refactor `fs.copy` in v3 anyway.

althonos (approved):

althonos:
Thank you! First PR out of the way :smiley:

PR #462, opened 2021-03-20
Added support for MDTM command in FTPFS

### LINKED ISSUE #456: FTPFS should use MDTM
Regarding fs.ftpfs.FS.getinfo (https://github.com/PyFilesystem/pyfilesystem2/blob/64d7a52e1f3037f33937f91e9db812cafdbbafc0/fs/ftpfs.py#L648)

That function tries to use MLST to retrieve arbitrary information of a single file and defaults to LIST if MLST is not supported. 
This makes `fs.copy_dir_if_newer` incredibly inefficient on systems that do not support MLST. 

Imo, an easy solution would be to possibly support MDTM as a fallback if only the modified datetime is requested, e.g. if `namespace == ("detail", "modified")`.
Another, probably more complex improvement, would be to cache the LIST response rather than call it for every single file.

I can submit a PR for the first suggestion, if I get the okay that this is something that should be included. Possibly also the second, although I'm not sure yet how to best implement that.

atollk:
## Type of changes

- Other

## Checklist

## Description

Fixes #456

Added a `getmodified` function to `fs.base.FS` which defaults to calling `getinfo`. In `FTPFS`, a check for the support of `MDTM` was added to provide a more efficient command for this specialized request.

Two calls to `FS.getinfo(...).modified` in `fs.copy` were replaced by calls to `FS.getmodified`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `fs.copy.copy_file_if`, `fs.copy.copy_dir_if`, and `fs.copy.copy_fs_if`. 
   Closes [#458](https://github.com/PyFilesystem/pyfilesystem2/issues/458).
+- Added `fs.base.FS.getmodified`.
   
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `fs.copy.copy_file_if`, `fs.copy.copy_dir_if`, and `fs.copy.copy_fs_if`. 
   Closes [#458](https://github.com/PyFilesystem/pyfilesystem2/issues/458).
   
+### Changed
+
+- FTP servers that do not support the MLST command now try to use the MDTM command to
+  retrieve the last modification timestamp of a resource. 
+  Closes [#456](https://github.com/PyFilesystem/pyfilesystem2/pull/456).
+  
 ### Fixed
 
 - Fixed performance bugs in `fs.copy.copy_dir_if_newer`. Test cases were adapted to catch those bugs in the future.
--- a/docs/source/interface.rst
+++ b/docs/source/interface.rst
@@ -20,6 +20,7 @@ The following is a complete list of methods on PyFilesystem objects.
 * :meth:`~fs.base.FS.getdetails` Get details info namespace for a resource.
 * :meth:`~fs.base.FS.getinfo` Get info regarding a file or directory.
 * :meth:`~fs.base.FS.getmeta` Get meta information for a resource.
+* :meth:`~fs.base.FS.getmodified` Get info regarding the last modified time of a resource.
 * :meth:`~fs.base.FS.getospath` Get path with encoding expected by the OS.
 * :meth:`~fs.base.FS.getsize` Get the size of a file.
 * :meth:`~fs.base.FS.getsyspath` Get the system path of a resource, if one exists.
--- a/fs/base.py
+++ b/fs/base.py
@@ -712,8 +712,6 @@ class FS(object):
         it might only have limited accuracy.
 
         """
-        if self.getmeta().get("supports_mtime", False):
-            return self.getinfo(path, namespaces=["modified"]).modified
         return self.getinfo(path, namespaces=["details"]).modified
 
     def getmeta(self, namespace="standard"):
@@ -753,8 +751,6 @@ class FS(object):
         read_only           `True` if this filesystem is read only.
         supports_rename     `True` if this filesystem supports an
                             `os.rename` operation.
-        supports_mtime      `True` if this filesystem supports a native
-                            operation to retrieve the "last modified" time.
         =================== ============================================
 
         Most builtin filesystems will provide all these keys, and third-
--- a/fs/base.py
+++ b/fs/base.py
@@ -754,7 +754,7 @@ class FS(object):
         supports_rename     `True` if this filesystem supports an
                             `os.rename` operation.
         supports_mtime      `True` if this filesystem supports a native
-                            operation to retreive the "last modified" time.
+                            operation to retrieve the "last modified" time.
         =================== ============================================
 
         Most builtin filesystems will provide all these keys, and third-
--- a/fs/base.py
+++ b/fs/base.py
@@ -697,6 +697,25 @@ class FS(object):
 
     gettext = _new_name(readtext, "gettext")
 
+    def getmodified(self, path):
+        # type: (Text) -> Optional[datetime]
+        """Get the timestamp of the last modifying access of a resource.
+
+        Arguments:
+            path (str): A path to a resource.
+
+        Returns:
+            datetime: The timestamp of the last modification.
+
+        The *modified timestamp* of a file is the point in time
+        that the file was last changed. Depending on the file system,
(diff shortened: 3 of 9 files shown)


atollk:
Not sure how I would add tests for this. Are there any tests that run with an actual FTP server? Should I do extensive mocking to simulate a server that supports MDTM?

atollk:
I added a test to `test_ftpfs` which, I think, decently covers this new feature, so I consider this PR ready to be reviewed / merged now.

fs/ftpfs.py
```diff
@@ -667,6 +668,18 @@ def getinfo(self, path, namespaces=None):
                 }
             )
 
+        if "modified" in namespaces:
+            if "basic" in namespaces or "details" in namespaces:
+                raise ValueError(
+                    'Cannot use the "modified" namespace in combination with others.'
+                )
```
althonos, line 675: Why would you block `modified` from being used with the other namespaces? In particular, `basic` is always implied, so you can't really return an `Info` without the basic namespace.

In case of both `modified` and `details` being present, we could at most raise a warning, but the code should still work.

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `fs.copy.copy_file_if`, `fs.copy.copy_dir_if`, and `fs.copy.copy_fs_if`. 
   Closes [#458](https://github.com/PyFilesystem/pyfilesystem2/issues/458).
```
althonos, line 15: Could you add an entry to the **Added** section to note the `FS.getmodified` method was introduced here? Also, please add the `FS.getmodified` method to the list in `docs/source/interface.rst`.

fs/base.py
```diff
@@ -734,6 +753,8 @@ def getmeta(self, namespace="standard"):
         read_only           `True` if this filesystem is read only.
         supports_rename     `True` if this filesystem supports an
                             `os.rename` operation.
+        supports_mtime      `True` if this filesystem supports a native
+                            operation to retreive the "last modified" time.
```
althonos, line 757: retreive :arrow_right:  retrieve

tests/test_memoryfs.py
```diff
@@ -72,14 +72,13 @@ def test_copy_preserve_time(self):
         self.fs.makedir("bar")
         self.fs.touch("foo/file.txt")
 
-        namespaces = ("details", "modified")
-        src_info = self.fs.getinfo("foo/file.txt", namespaces)
+        src_info = self.fs.getmodified("foo/file.txt")
```
althonos, line 75: This should be renamed to `src_datetime` since it's not an info anymore for clarity.

tests/test_memoryfs.py
```diff
@@ -72,14 +72,13 @@ def test_copy_preserve_time(self):
...
 
         self.fs.copy("foo/file.txt", "bar/file.txt", preserve_time=True)
         self.assertTrue(self.fs.exists("bar/file.txt"))
 
-        dst_info = self.fs.getinfo("bar/file.txt", namespaces)
-        self.assertEqual(dst_info.modified, src_info.modified)
+        dst_info = self.fs.getmodified("bar/file.txt")
```
althonos, line 80: Same as earlier, `dst_datetime`.

fs/info.py
```diff
@@ -317,9 +317,12 @@ def modified(self):
...
         """
-        self._require_namespace("details")
-        _time = self._make_datetime(self.get("details", "modified"))
-        return _time
+        try:
+            self._require_namespace("details")
+            return self._make_datetime(self.get("details", "modified"))
+        except MissingInfoNamespace:
+            self._require_namespace("modified")
+            return self._make_datetime(self.get("modified", "modified"))
```
althonos, line 325: Since the `MissingInfoNamespace` is actually raised by `self._require_namespace(...)`, I think it would be nicer to do the following (minimal branching, and at most one exception raised):

```python
@property
def modified(self):
    namespace =  "modified" if "modified" in self.raw else "details"
    self._require_namespace(namespace)
    return self._make_datetime(self.get(namespace, "modified"))
```

althonos (changes requested):
Great work! A few fixes here and there, but the global structure is as intended :smile:

fs/ftpfs.py
atollk: I added that because I feel like requesting both "details" and "modified" at the same time is never a good idea to do intentionally. But I see your point; I'll change it to a warning.

fs/info.py
atollk: I don't feel like that's a good change just to get rid of one possible `raise`. That would make `modified` the only property in `Info` to not use `get` but access `raw` instead, reducing encapsulation of that class.

althonos: Then I guess using `has_namespace` instead, which is a public method, would be fine:
```python
@property
def modified(self):
    namespace =  "modified" if self.has_namespace("modified") else "details"
    self._require_namespace(namespace)
    return self._make_datetime(self.get(namespace, "modified"))
```

atollk: Done.

atollk:
Hi. Any update on this?

althonos:
@atollk : sorry for the delay, I had a very intense month but I have a bit more time now. 

I changed the current implementation a bit because the `Info.modified` approach had one caveat: it would cause an `Info` object to be created without the `basic` namespace, which it should normally always have, so I was a bit worried about the rest of the interface and that's why I didn't merge right away.

Instead, I kept the `FS.getmodified` method, and simply overloaded it in `FTPFS`. This means that code using `getmodified` should be faster now, and since you updated the relevant functions in `fs.copy` I think this will successfully fix #456 nevertheless.

althonos (approved):

althonos:
@atollk : All clear now. I'll leave you time to object if there is anything to change, ping me when you're done reading and I'll merge.

tests/test_ftpfs.py
```diff
@@ -265,7 +265,8 @@ def test_getinfo_modified(self):
         self.assertIn("MDTM", self.fs.features)
         self.fs.create("bar")
         mtime_detail = self.fs.getinfo("bar", ("basic", "details")).modified
-        mtime_modified = self.fs.getinfo("bar", ("modified",)).modified
+        mtime_modified = self.fs.getmodified("bar")
+        print(mtime_detail, mtime_modified)
```
atollk, line 269: Is this `print` supposed to stay or be removed?

atollk:
Except for that one comment I made it looks good to me (y)

tests/test_ftpfs.py
althonos: Ahah, good catch, thanks!

althonos:
Perfect, merging then! Sorry for the delay.

docs/source/interface.rst
```diff
@@ -20,6 +20,7 @@ The following is a complete list of methods on PyFilesystem objects.
 * :meth:`~fs.base.FS.getdetails` Get details info namespace for a resource.
 * :meth:`~fs.base.FS.getinfo` Get info regarding a file or directory.
 * :meth:`~fs.base.FS.getmeta` Get meta information for a resource.
+* :meth:`~fs.base.FS.getmodified` Get info regarding the last modified time of a resource.
```
lurch, line 23: Should this have been "Get the last modified time of a resource."? The "info regarding" seems superfluous?

althonos: @lurch : Good catch, I think I'm due for a documentation PR next.

PR #463, opened 2021-03-20
Added preserve_time parameter to move, copy, and mirror

### LINKED ISSUE #450: original timestamp (ctime, mtime) cannot survive a copy_fs or mirror
as long as there's actual "copy" operation from one fs to another, then those "touched" files will lost their original ctime or mtime, which would be updated to the time when that copy operation happened.
this is not expected feature. though calling it "critical bug" might be too harsh but still it's much better if it could be fixed.

atollk:
## Type of changes

- New feature

## Checklist

## Description

Fixes #450 by adding a `preserve_time` parameter to all relevant functions. The parameter is `False` by default to retain previous behaviour. However, it could be argued that the default should be `True` to mimic more common behavior found in operating systems.

### DIFF
--- a/fs/_bulk.py
+++ b/fs/_bulk.py
@@ -56,20 +56,30 @@ class _Task(object):
 class _CopyTask(_Task):
     """A callable that copies from one file another."""
 
-    def __init__(self, src_file, dst_file):
-        # type: (IO, IO) -> None
-        self.src_file = src_file
-        self.dst_file = dst_file
+    def __init__(
+        self,
+        src_fs,  # type: FS
+        src_path,  # type: Text
+        dst_fs,  # type: FS
+        dst_path,  # type: Text
+        preserve_time,  # type: bool
+    ):
+        # type: (...) -> None
+        self.src_fs = src_fs
+        self.src_path = src_path
+        self.dst_fs = dst_fs
+        self.dst_path = dst_path
+        self.preserve_time = preserve_time
 
     def __call__(self):
         # type: () -> None
-        try:
-            copy_file_data(self.src_file, self.dst_file, chunk_size=1024 * 1024)
-        finally:
-            try:
-                self.src_file.close()
-            finally:
-                self.dst_file.close()
+        copy_file_internal(
+            self.src_fs,
+            self.src_path,
+            self.dst_fs,
+            self.dst_path,
+            preserve_time=self.preserve_time,
+        )
 
 
 class Copier(object):
@@ -88,7 +98,7 @@ class Copier(object):
     def start(self):
         """Start the workers."""
         if self.num_workers:
-            self.queue = Queue(maxsize=self.num_workers)
+            self.queue = Queue()
             self.workers = [_Worker(self) for _ in range(self.num_workers)]
             for worker in self.workers:
                 worker.start()
@@ -133,12 +143,5 @@ class Copier(object):
                 src_fs, src_path, dst_fs, dst_path, preserve_time=preserve_time
             )
         else:
-            # TODO(preserve_time)
-            src_file = src_fs.openbin(src_path, "r")
-            try:
-                dst_file = dst_fs.openbin(dst_path, "w")
-            except Exception:
-                src_file.close()
-                raise
-            task = _CopyTask(src_file, dst_file)
+            task = _CopyTask(src_fs, src_path, dst_fs, dst_path, preserve_time)
             self.queue.put(task)
--- a/fs/_bulk.py
+++ b/fs/_bulk.py
@@ -13,12 +13,11 @@ from six.moves.queue import Queue
 
 from .copy import copy_file_internal
 from .errors import BulkCopyFailed
-from .tools import copy_file_data
 
 if typing.TYPE_CHECKING:
     from .base import FS
     from types import TracebackType
-    from typing import IO, List, Optional, Text, Type
+    from typing import List, Optional, Text, Type
 
 
 class _Worker(threading.Thread):
--- a/fs/_bulk.py
+++ b/fs/_bulk.py
@@ -11,13 +11,14 @@ import typing
 
 from six.moves.queue import Queue
 
-from .copy import copy_file_internal
+from .copy import copy_file_internal, copy_modified_time
 from .errors import BulkCopyFailed
+from .tools import copy_file_data
 
 if typing.TYPE_CHECKING:
     from .base import FS
     from types import TracebackType
-    from typing import List, Optional, Text, Type
+    from typing import List, Optional, Text, Type, IO, Tuple
 
 
 class _Worker(threading.Thread):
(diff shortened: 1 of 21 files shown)


atollk:
I wanted to add https://github.com/wolever/parameterized as a dependency because parameterization was done already anyway, just in a hacky self-implemented way. I don't actually get how to add test dependencies though?

althonos:
If you want to add a dependency for the tests, add it to `tests/requirements.txt`, pinning the version with `~=` if possible. It will get installed by `tox` and by the CI runner.

atollk:
@althonos From what I can tell, this doesn't work [here](https://ci.appveyor.com/project/willmcgugan/pyfilesystem2/build/job/d6v8sh072h8e27ni), as the tests are run without tox.

althonos:
@atollk : that was not expected, I patched `appveyor.yml`. I need to refactor the CI on AppVeyor anyway, since we are currently not testing all supported versions or measuring coverage there.

atollk:
I just saw that `FTPFS.setinfo` actually doesn't do anything, so I also added support for the MFMT command so that `preserve_time` works when copying to a FTPFS too.

fs/_bulk.py
```diff
@@ -88,7 +97,7 @@ def __init__(self, num_workers=4):
     def start(self):
         """Start the workers."""
         if self.num_workers:
-            self.queue = Queue(maxsize=self.num_workers)
+            self.queue = Queue()
```
althonos, line 100: Why the change here?

fs/base.py
```diff
@@ -417,16 +426,25 @@ def copy(self, src_path, dst_path, overwrite=False):
             with closing(self.open(src_path, "rb")) as read_file:
                 # FIXME(@althonos): typing complains because open return IO
                 self.upload(dst_path, read_file)  # type: ignore
+            copy_mtime(self, src_path, self, dst_path)
```
althonos, line 429: Shouldn't it be:
```python
if preserve_time:
    copy_mtime(...)
```
here?

fs/base.py
```diff
@@ -1133,6 +1155,7 @@ def move(self, src_path, dst_path, overwrite=False):
             with self.open(src_path, "rb") as read_file:
                 # FIXME(@althonos): typing complains because open return IO
                 self.upload(dst_path, read_file)  # type: ignore
+            copy_mtime(self, src_path, self, dst_path)
```
althonos, line 1158: Same here:
```python
if preserve_time:
    copy_mtime(...)
```

althonos: Also, what happens if both resources have a syspath? There is an early return above, so I don't think the *mtime* is copied in that case.

fs/copy.py
```diff
@@ -286,9 +334,10 @@ def dst():
     from ._bulk import Copier
 
     with src() as _src_fs, dst() as _dst_fs:
-        with _src_fs.lock(), _dst_fs.lock():
-            _thread_safe = is_thread_safe(_src_fs, _dst_fs)
-            with Copier(num_workers=workers if _thread_safe else 0) as copier:
+        _thread_safe = is_thread_safe(_src_fs, _dst_fs)
+        copier = Copier(num_workers=workers if _thread_safe else 0)
+        with copier:
+            with _src_fs.lock(), _dst_fs.lock():
```
althonos, line 340: Any reason to create the copier before locking instead of after like before?

fs/memoryfs.py
```diff
@@ -444,7 +444,7 @@ def makedir(
                 parent_dir.set_entry(dir_name, new_dir)
             return self.opendir(path)
 
-    def move(self, src_path, dst_path, overwrite=False):
+    def move(self, src_path, dst_path, overwrite=False, preserve_time=False):
```
althonos, line 447: `MemoryFS` actually supports created/modified times, so you should be able to get it to preserve times here too.

fs/wrapfs.py
```diff
@@ -167,18 +167,18 @@ def makedir(
...
         dst_fs, _dst_path = self.delegate_path(dst_path)
         with unwrap_errors({_src_path: src_path, _dst_path: dst_path}):
             if not overwrite and dst_fs.exists(_dst_path):
                 raise errors.DestinationExists(_dst_path)
-            move_file(src_fs, _src_path, dst_fs, _dst_path)
+            move_file(src_fs, _src_path, dst_fs, _dst_path, preserve_time=preserve_time)
 
-    def movedir(self, src_path, dst_path, create=False):
-        # type: (Text, Text, bool) -> None
+    def movedir(self, src_path, dst_path, create=False, preserve_time=False):
```
althonos, line 180: I think you're not using `preserve_time` in the subsequent `move_dir` call here.

althonos (changes requested):
I'm not 100% satisfied with the `copy_mtime`  name, because we never refer to `mtime` anywhere else in the API (we use `modified` everywhere). Yet `copy_modified` does not convey the function usecase very well. `copy_modified_time` maybe? Might be a bit too long.

There are just a few minore issues here and there but overall this looks fine.

fs/base.py
atollk: I think `os.rename` doesn't actually touch the mtime value, at least on UNIX and Windows. I added the check anyway, just to be safe.

fs/copy.py
atollk: I had to change the logic of `Copier` because of some internal locking mechanics. But I just noticed that that change might actually have caused all workers to work sequentially, rather than in parallel. So I reverted most of my changes in `fs._bulk`. The modification times are now updated when closing the `Copier` instance, after all tasks have finished.

fs/_bulk.py
atollk: see my other comment [here](https://github.com/PyFilesystem/pyfilesystem2/pull/463#discussion_r609513389)

althonos (approved):
LGTM, congrats!

althonos:
I merged this one because I have already been too nit-picky here, but for the remaining PRs, would you be able to rebase them to `master` so that the commit history is still a bit organized?

atollk:
@althonos Sure can do. But I'd also be happy to write a more thorough commit message for a squash merge, since I feel a good part of the commits in those PRs are just one-file fixes anyway. Or do you not do squashes in this repo?

althonos:
The developer interaction was minimal here for some time so we just went with merges. I do squashes in other projects but for PyFilesystem I didn't wanna change what was there so branch/merges are fine. As long as they are properly rebased, the network is still fine: https://github.com/PyFilesystem/pyfilesystem2/network

atollk:
@althonos In case you didn't see, I did a rebase in both https://github.com/PyFilesystem/pyfilesystem2/pull/462 and https://github.com/PyFilesystem/pyfilesystem2/pull/464.

althonos:
Thanks, perfect, i'll have a look tomorrow !

PR #464, opened 2021-03-22
Add `filter_glob` and `exclude_glob` to fs.walk

### LINKED ISSUE #371: filter_dirs in walk.dirs behaves unintuitive and its description is a bit misleading
Talking about https://docs.pyfilesystem.org/en/latest/reference/walk.html#fs.walk.BoundWalker.dirs and maybe `walk.files` is affected as well.

The documentation talks about (emphasis by me):

> **dirs(path='/', \*\*kwargs)**
> Walk a filesystem, yielding **ABSOLUTE** paths to directories.

and

> **filter_dirs** _(list, optional)_ – A list of patterns that will be used to match directories **PATHS**. The walk will only open directories that match at least one of these patterns.

This gives me the impression that the filters are applied to the absolute paths.

While in fact, if I understood and investigated correctly, these patterns match the directory NAMES. Here's an example to show that behavior:
```python
with MemoryFS() as myfs:
    myfs.makedirs('/foo/bar')
    print('-------------------')
    for path in myfs.walk.dirs():
        print(path)
    print('-------------------')
    for path in myfs.walk.dirs(filter_dirs=['*bar*']):
        print(path)
    print('-------------------')
```
Actual output:
```
-------------------
/foo
/foo/bar
-------------------
-------------------
```
Expected output:
```
-------------------
/foo
/foo/bar
-------------------
/foo/bar
-------------------
```

This is probably not a bug in the code. However, the documentation is misleading and my question would be how I can get to my desired behavior.

### LINKED ISSUE #459: Add a global filter to fs.walk.Walker
`fs.walk.Walker` can take `filter` and `exclude` parameters to include/exclude certain file base names, as well as `filter_dirs` and `exclude_dirs` parameters to include/exclude entire directories.

My suggestion is to introduce two new parameters, `filter_glob` and `exclude_glob`, to work on arbitrary path objects. This would make working with filters mixed of both directories and files much easier.

For example, consider a file system with three files:
```
foo/a.py
foo/b.py
bar/a.py
```

Walking over all files except for `foo/a.py` is, afaik, impossible with the current `Walker` functionality. My suggestion would allow for:

```python
fs.walk.Walker(filter_glob=["/"], exclude_glob=["/foo/a.py"])
```

atollk:
## Type of changes

- New feature
- Documentation / docstrings

## Checklist

## Description

Fixes #371. Fixes #459 .
Added parameters `filter_glob` and `exclude_glob` to `fs.walk.Walker` to perform selection on the walked resources based on their global path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- To `fs.walk.Walker`, added parameters `filter_glob` and `exclude_glob`.
+- Added `filter_glob` and `exclude_glob` parameters to `fs.walk.Walker`.
   Closes [#459](https://github.com/PyFilesystem/pyfilesystem2/issues/459).
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+
+- To `fs.walk.Walker`, added parameters `filter_glob` and `exclude_glob`.
+  Closes [#459](https://github.com/PyFilesystem/pyfilesystem2/issues/459).
+
+### Fixed
+- Elaborated documentation of `filter_dirs` and `exclude_dirs` in `fs.walk.Walker`.
+  Closes [#371](https://github.com/PyFilesystem/pyfilesystem2/issues/371).
+
 
 ## [2.4.16] - 2022-05-02
 
--- a/fs/base.py
+++ b/fs/base.py
@@ -21,7 +21,7 @@ import warnings
 from contextlib import closing
 from functools import partial, wraps
 
-from . import copy, errors, fsencode, iotools, tools, walk, wildcard
+from . import copy, errors, fsencode, iotools, tools, walk, wildcard, glob
 from .copy import copy_modified_time
 from .glob import BoundGlobber
 from .mode import validate_open_mode
@@ -1653,8 +1653,8 @@ class FS(object):
         if self.isclosed():
             raise errors.FilesystemClosed()
 
-    def match(self, patterns, name):
-        # type: (Optional[Iterable[Text]], Text) -> bool
+    def match(self, patterns, name, accept_prefix=False):
+        # type: (Optional[Iterable[Text]], Text, bool) -> bool
         """Check if a name matches any of a list of wildcards.
 
         If a filesystem is case *insensitive* (such as Windows) then
@@ -1695,6 +1695,61 @@ class FS(object):
         )
         matcher = wildcard.get_matcher(patterns, case_sensitive)
         return matcher(name)
+
+    def match_glob(self, patterns, path, accept_prefix=False):
+        # type: (Optional[Iterable[Text]], Text, bool) -> bool
+        """Check if a path matches any of a list of glob patterns.
+
+        If a filesystem is case *insensitive* (such as Windows) then
+        this method will perform a case insensitive match (i.e. ``*.py``
+        will match the same names as ``*.PY``). Otherwise the match will
+        be case sensitive (``*.py`` and ``*.PY`` will match different
+        names).
+
+        Arguments:
+            patterns (list, optional): A list of patterns, e.g.
+                ``['*.py']``, or `None` to match everything.
+            path (str): A resource path, starting with "/".
+            accept_prefix (bool): If ``True``, the path is
+                not required to match the wildcards themselves
+                but only need to be a prefix of a string that does.
+
+        Returns:
+            bool: `True` if ``path`` matches any of the patterns.
+
+        Raises:
+            TypeError: If ``patterns`` is a single string instead of
+                a list (or `None`).
+            ValueError: If ``path`` is not a string starting with "/".
+
+        Example:
+            >>> my_fs.match_glob(['*.py'], '/__init__.py')
+            True
+            >>> my_fs.match_glob(['*.jpg', '*.png'], '/foo.gif')
+            False
+            >>> my_fs.match_glob(['dir/file.txt'], '/dir/', accept_prefix=True)
+            True
+            >>> my_fs.match_glob(['dir/file.txt'], '/dir/gile.txt', accept_prefix=True)
+            False
+
+        Note:
+            If ``patterns`` is `None` (or ``['*']``), then this
+            method will always return `True`.
+
+        """
+        if patterns is None:
+            return True
+        if not path or path[0] != "/":
+            raise ValueError("%s needs to be a string starting with /" % path)
+        if isinstance(patterns, six.text_type):
+            raise TypeError("patterns must be a list or sequence")
+        case_sensitive = not typing.cast(
+            bool, self.getmeta().get("case_insensitive", False)
+        )
+        matcher = glob.get_matcher(
+            patterns, case_sensitive, accept_prefix=accept_prefix
+        )
+        return matcher(path)
 
     def tree(self, **kwargs):
         # type: (**Any) -> None
--- a/fs/base.py
+++ b/fs/base.py
@@ -1653,8 +1653,8 @@ class FS(object):
         if self.isclosed():
             raise errors.FilesystemClosed()
 
-    def match(self, patterns, name, accept_prefix=False):
-        # type: (Optional[Iterable[Text]], Text, bool) -> bool
+    def match(self, patterns, name):
+        # type: (Optional[Iterable[Text]], Text) -> bool
         """Check if a name matches any of a list of wildcards.
 
         If a filesystem is case *insensitive* (such as Windows) then
@@ -1711,7 +1711,7 @@ class FS(object):
                 ``['*.py']``, or `None` to match everything.
             path (str): A resource path, starting with "/".
             accept_prefix (bool): If ``True``, the path is
-                not required to match the wildcards themselves
+                not required to match the patterns themselves
                 but only need to be a prefix of a string that does.
 
         Returns:
@@ -1729,6 +1729,8 @@ class FS(object):
             False
             >>> my_fs.match_glob(['dir/file.txt'], '/dir/', accept_prefix=True)
             True
+            >>> my_fs.match_glob(['dir/file.txt'], '/dir/', accept_prefix=False)
+            False
             >>> my_fs.match_glob(['dir/file.txt'], '/dir/gile.txt', accept_prefix=True)
             False
 
--- a/fs/glob.py
+++ b/fs/glob.py
@@ -4,22 +4,31 @@
 from __future__ import unicode_literals
 
 import typing
+from functools import partial
 
 import re
 from collections import namedtuple
 
-from . import wildcard
 from ._repr import make_repr
 from .lrucache import LRUCache
 from .path import iteratepath
+
 
 GlobMatch = namedtuple("GlobMatch", ["path", "info"])
 Counts = namedtuple("Counts", ["files", "directories", "data"])
 LineCounts = namedtuple("LineCounts", ["lines", "non_blank"])
 
 if typing.TYPE_CHECKING:
-    from typing import Iterator, List, Optional, Pattern, Text, Tuple
-
+    from typing import (
+        Iterator,
+        List,
+        Optional,
+        Pattern,
+        Text,
+        Tuple,
+        Iterable,
+        Callable,
+    )
     from .base import FS
 
 
@@ -28,17 +37,87 @@ _PATTERN_CACHE = LRUCache(
 )  # type: LRUCache[Tuple[Text, bool], Tuple[int, bool, Pattern]]
 
 
+def _split_pattern_by_rec(pattern):
+    # type: (Text) -> List[Text]
+    """Split a glob pattern at its directory seperators (/).
+
+    Takes into account escaped cases like [/].
+    """
+    indices = [-1]
+    bracket_open = False
+    for i, c in enumerate(pattern):
+        if c == "/" and not bracket_open:
+            indices.append(i)
+        elif c == "[":
+            bracket_open = True
+        elif c == "]":
+            bracket_open = False
+
+    indices.append(len(pattern))
+    return [pattern[i + 1 : j] for i, j in zip(indices[:-1], indices[1:])]
+
+
+def _translate(pattern, case_sensitive=True):
+    # type: (Text, bool) -> Text
+    """Translate a wildcard pattern to a regular expression.
+
+    There is no way to quote meta-characters.
+    Arguments:
+        pattern (str): A wildcard pattern.
+        case_sensitive (bool): Set to `False` to use a case
+            insensitive regex (default `True`).
+
+    Returns:
+        str: A regex equivalent to the given pattern.
+
+    """
+    if not case_sensitive:
+        pattern = pattern.lower()
+    i, n = 0, len(pattern)
+    res = []
+    while i < n:
+        c = pattern[i]
+        i = i + 1
+        if c == "*":
+            res.append("[^/]*")
+        elif c == "?":
+            res.append("[^/]")
+        elif c == "[":
+            j = i
+            if j < n and pattern[j] == "!":
+                j = j + 1
+            if j < n and pattern[j] == "]":
+                j = j + 1
+            while j < n and pattern[j] != "]":
+                j = j + 1
+            if j >= n:
+                res.append("\\[")
+            else:
+                stuff = pattern[i:j].replace("\\", "\\\\")
+                i = j + 1
+                if stuff[0] == "!":
+                    stuff = "^" + stuff[1:]
+                elif stuff[0] == "^":
+                    stuff = "\\" + stuff
+                res.append("[%s]" % stuff)
+        else:
+            res.append(re.escape(c))
+    return "".join(res)
+
+
 def _translate_glob(pattern, case_sensitive=True):
     levels = 0
     recursive = False
     re_patterns = [""]
     for component in iteratepath(pattern):
-        if component == "**":
-            re_patterns.append(".*/?")
+        if "**" in component:
             recursive = True
+            split = component.split("**")
+            split_re = [_translate(s, case_sensitive=case_sensitive) for s in split]
+            re_patterns.append("/?" + ".*/?".join(split_re))
         else:
             re_patterns.append(
-                "/" + wildcard._translate(component, case_sensitive=case_sensitive)
+                "/" + _translate(component, case_sensitive=case_sensitive)
             )
         levels += 1
     re_glob = "(?ms)^" + "".join(re_patterns) + ("/$" if pattern.endswith("/") else "$")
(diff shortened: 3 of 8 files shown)


fs/wildcard.py
```diff
@@ -123,55 +135,92 @@ def get_matcher(patterns, case_sensitive):
...
-def _translate(pattern, case_sensitive=True):
-    # type: (Text, bool) -> Text
-    """Translate a wildcard pattern to a regular expression.
+class _PatternMatcherToken:
+    def __init__(self, token_str):
+        self.token_str = token_str
 
-    There is no way to quote meta-characters.
+    def __repr__(self):
+        return "_PatternMatcherToken (" + self.token_str + ")"
```
althonos, line 203: Use a repr string here:
```python
return "_PatternMatcherToken({!r})".format(self.token_str)
```
this will make `repr` roundtrip with `eval`, which is good practice for simple objects.

fs/wildcard.py
```diff
@@ -123,55 +135,92 @@ def get_matcher(patterns, case_sensitive):
...
     else:
-        return partial(imatch_any, patterns)
+        return lambda name: imatch_any(patterns, name, accept_prefix)
+
+
+class _PatternMatcher:
+    def __init__(self, pattern, case_sensitive, accept_prefixes):
+        self.case_sensitive = case_sensitive
+        self.accept_prefixes = accept_prefixes
+
```
althonos, line 147: I'd also store the pattern for good measure as an attribute here, it could help debugging.

althonos: Also, this would allow writing a roundtripping `__repr__` implementation:

```python
from .repr import make_repr

class _PatternMatcher:
    def __repr__(self):
        return make_repr(
            '_PatternMatcher', 
            self.pattern, 
            case_sensitive=(self.case_sentitive, None), 
            accept_prefixes=(self.accept_prefixes, None)
        )
```

fs/base.py
```diff
@@ -1645,6 +1645,9 @@ def match(self, patterns, name):
             patterns (list, optional): A list of patterns, e.g.
                 ``['*.py']``, or `None` to match everything.
             name (str): A file or directory name (not a path)
+            accept_prefix (bool): If ``True``, the name is
```
althonos, line 1648: Could you maybe add an example to the **Example** section below showing what the `accept_prefix` does? I think it'd be easier to understand that way.

fs/wildcard.py
```diff
@@ -123,55 +135,92 @@ def get_matcher(patterns, case_sensitive):
...
+                while (asterik_len + i) < len(pattern) and pattern[
+                    asterik_len + i
+                ] == "*":
+                    asterik_len += 1
+                if asterik_len > 2:
+                    raise ValueError(
+                        pattern
+                        + " is not a valid wildcard pattern. "
+                        + "(sequence of three or more *)"
+                    )
```
althonos, line 173: Seeing these different errors, I'm tempted to add a `PatternError` to `fs.errors` here, and maybe subclass it (either `TooManyTokensError` or `UnmatchedTokenError`).

althonos (changes requested):
Nice job once again :wink: 

My largest concern is about whether it would be possible to remove the `_translate` function completely, and use `_PatternMatcher` in `fs.glob` as well to reduce code duplication. Otherwise, it's mostly smaller stuff.

I'll have a closer look at the actual files at some point because GitHub diffs are really not helping here, but the first read didn't raise any red flag :smile:

fs/glob.py
```diff
@@ -27,6 +26,54 @@
 )  # type: LRUCache[Tuple[Text, bool], Tuple[int, bool, Pattern]]
 
 
+def _translate(pattern, case_sensitive=True):
```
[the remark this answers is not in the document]
atollk: Yes, pretty much, but not "moved" and rather "copied". 
I think it would be possible to just merge the two; at least, I don't see any difference in their feature set or behavior anymore. 

The only odd point I can find (from [here](https://docs.pyfilesystem.org/en/latest/globbing.html)) is:
> *.py?
> Matches all .py files and .pyi, .pyc etc in the currenct directory.

That's not the way [UNIX glob](https://man7.org/linux/man-pages/man7/glob.7.html) works (which this syntax is based on, I assume) or the way `fs.wildcard` works, nor is this actually tested anywhere. So it might just be an error in documentation.

fs/wildcard.py
```diff
@@ -123,55 +135,92 @@ def get_matcher(patterns, case_sensitive):
     if not patterns:
         return lambda name: True
     if case_sensitive:
-        return partial(match_any, patterns)
+        return lambda name: match_any(patterns, name, accept_prefix)
```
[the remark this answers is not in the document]
atollk: Is that desired though? To my knowledge, lambdas are in general considered preferable over `functools` nowadays.

fs/wildcard.py
atollk: Thanks. TIL

fs/wildcard.py
atollk: I added a `PatternError` which also stores the position at which the error occurred. That seemed nicer to me than having a new class for each type of error but tell me if that isn't what you had in mind.

fs/glob.py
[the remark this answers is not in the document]
althonos: Yes, I think it's a documentation error too. The current implementation just replaces a `?` in the wildcard with a `.` in the corresponding regex.

fs/wildcard.py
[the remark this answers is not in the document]
althonos: I'd generally go for the opposite, for the simple reason that `lambda` functions cannot be pickled. This means there are places where you can't use them (for instance as a callback for a background thread), whereas a `partial` will work fine. 

`functools.partial` also gives you a nicer `repr` string compared to a lambda IMO.

althonos (changes requested):

atollk:
Alright, so, after taking a step back and having a look at the larger picture again, it made more sense to me to put all that logic into `fs.glob` instead of `fs.wildcard`, so I reverted the change in the latter. I also don't think the two can (or should) be merged anymore, due to the specific behavior of glob. Most notably `/**/` matches `/`, which is not what you would expect from a wildcard.

Either way, several of your previous comments became obsolete through this and I implemented the rest (like `partial` instead of `lambda`).

atollk:
Hi. Any update on this?

atollk:
I just rebased this branch onto the newest `master`, so it's ready to merge now imo.

althonos:
@atollk : I think you rebased against your old `master` without pulling from PyFilesystem/pyfilesystem2 first, the code still have merges from a prior PR.

atollk:
@althonos Should be fine now.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- To `fs.walk.Walker`, added parameters `filter_glob` and `exclude_glob`.
```
lurch, line 13: I think "- Added `filter_glob` and `exclude_glob` parameters to `fs.walk.Walker`." sounds more natural here.

fs/base.py
```diff
@@ -1648,8 +1648,8 @@ def check(self):
         if self.isclosed():
             raise errors.FilesystemClosed()
 
-    def match(self, patterns, name):
-        # type: (Optional[Iterable[Text]], Text) -> bool
+    def match(self, patterns, name, accept_prefix=False):
```
lurch, line 1651: Should `accept_prefix` be added to the "Arguments:" docstring here?

fs/base.py
```diff
@@ -1691,6 +1691,61 @@ def match(self, patterns, name):
...
+
+        Example:
+            >>> my_fs.match_glob(['*.py'], '/__init__.py')
+            True
+            >>> my_fs.match_glob(['*.jpg', '*.png'], '/foo.gif')
+            False
+            >>> my_fs.match_glob(['dir/file.txt'], '/dir/', accept_prefix=True)
+            True
+            >>> my_fs.match_glob(['dir/file.txt'], '/dir/gile.txt', accept_prefix=True)
+            False
```
lurch, line 1728: Perhaps it'd be helpful to show an example of how accept_prefix=True/False works differently? :shrug:

fs/base.py
```diff
@@ -1691,6 +1691,61 @@ def match(self, patterns, name):
...
+
+        Note:
+            If ``patterns`` is `None` (or ``['*']``), then this
+            method will always return `True`.
+
+        """
+        if patterns is None:
+            return True
+        if not path or path[0] != "/":
```
lurch, line 1737: I think this is the same as `if not path.startswith("/"):`

fs/base.py
```diff
@@ -1691,6 +1691,61 @@ def match(self, patterns, name):
...
+            raise ValueError("%s needs to be a string starting with /" % path)
+        if isinstance(patterns, six.text_type):
+            raise TypeError("patterns must be a list or sequence")
```
lurch, line 1740: This looks a bit weird - does it mean that passing e.g. `patterns=6` would be invalid, but not raise an error?

lurch: Oh, I see this is behaviour that you copied from `match()` just above :wink:

fs/base.py
```diff
@@ -1691,6 +1691,61 @@ def match(self, patterns, name):
...
+        will match the same names as ``*.PY``). Otherwise the match will
+        be case sensitive (``*.py`` and ``*.PY`` will match different
+        names).
+
+        Arguments:
+            patterns (list, optional): A list of patterns, e.g.
+                ``['*.py']``, or `None` to match everything.
+            path (str): A resource path, starting with "/".
+            accept_prefix (bool): If ``True``, the path is
+                not required to match the wildcards themselves
```
lurch, line 1709: Should "wildcards" here say "patterns" ?

fs/base.py
```diff
@@ -1691,6 +1691,61 @@ def match(self, patterns, name):
...
+
+        If a filesystem is case *insensitive* (such as Windows) then
+        this method will perform a case insensitive match (i.e. ``*.py``
+        will match the same names as ``*.PY``). Otherwise the match will
+        be case sensitive (``*.py`` and ``*.PY`` will match different
+        names).
+
+        Arguments:
+            patterns (list, optional): A list of patterns, e.g.
+                ``['*.py']``, or `None` to match everything.
```
lurch, line 1706: Looking at just the doc-strings, it's hard to see how `match_glob()` is different from `match()` ?

fs/base.py
lurch: Or was the addition of this extra parameter an error, as it doesn't actually seem to be used in this function?

fs/glob.py
```diff
@@ -27,17 +36,87 @@
 )  # type: LRUCache[Tuple[Text, bool], Tuple[int, bool, Pattern]]
 
 
+def _split_pattern_by_rec(pattern):
```
lurch, line 39: What does the `rec` in this function-name mean? I wonder if `_split_pattern_by_dir` or `_split_pattern_by_sep` would make more sense?

fs/glob.py
```diff
@@ -71,6 +150,8 @@ def match(pattern, path):
     except KeyError:
         levels, recursive, re_pattern = _translate_glob(pattern, case_sensitive=True)
         _PATTERN_CACHE[(pattern, True)] = (levels, recursive, re_pattern)
+    if path and path[0] != "/":
+        path = "/" + path
```
lurch, line 154: Is this the same as https://docs.pyfilesystem.org/en/latest/reference/path.html?#fs.path.abspath ?

fs/glob.py
```diff
@@ -91,9 +172,95 @@ def imatch(pattern, path):
...
+def match_any(patterns, path):
+    # type: (Iterable[Text], Text) -> bool
+    """Test if a path matches any of a list of patterns.
+
+    Will return `True` if ``patterns`` is an empty list.
+
+    Arguments:
+        patterns (list): A list of wildcard pattern, e.g ``["*.py",
+            "*.pyc"]``
+        name (str): A filename.
```
lurch, line 189: Says `name` here but `path` in the function definition. (Same "error" in `imatch_any` too.)

fs/glob.py
```diff
@@ -27,17 +36,87 @@
...
+        elif c == "[":
+            bracket_open = True
+        elif c == "]":
+            bracket_open = False
+
+    indices.append(len(pattern))
+    return [pattern[i + 1 : j] for i, j in zip(indices[:-1], indices[1:])]
+
+
+def _translate(pattern, case_sensitive=True):
```
lurch, line 59: The implementation of this function looks fairly hairy - would it make sense to add some unit-tests for it?

fs/glob.py
```diff
@@ -91,9 +172,95 @@ def imatch(pattern, path):
     except KeyError:
         levels, recursive, re_pattern = _translate_glob(pattern, case_sensitive=True)
```
lurch, line 173: Should this be `case_sensitive=False`? (If it does, I guess it shows how much thew unit-tests are lacking :wink: )

fs/glob.py
```diff
@@ -91,9 +172,95 @@ def imatch(pattern, path):
...
+    # type: (Iterable[Text], bool, bool) -> Callable[[Text], bool]
+    """Get a callable that matches paths against the given patterns.
+
+    Arguments:
+        patterns (list): A list of wildcard pattern. e.g. ``["*.py",
+            "*.pyc"]``
+        case_sensitive (bool): If ``True``, then the callable will be case
+            sensitive, otherwise it will be case insensitive.
+        accept_prefix (bool): If ``True``, the name is
+            not required to match the wildcards themselves
```
lurch, line 230: wildcards -> patterns ?

fs/walk.py
```diff
@@ -85,11 +87,22 @@ def __init__(
...
-                ``['*.svn', '*.git']``.
+                ``['*.svn', '*.git']``. Directories matching any of these
+                patterns will be removed from the walk.
             max_depth (int, optional): Maximum directory depth to walk.
+            filter_glob (list, optional): If supplied, this parameter
+                should be a list of path patterns e.g. ``["foo/**/*.py"]``.
+                Resources will only be returned if their global path or
+                an extension of it matches one of the patterns.
+            exclude_glob (list, optional): If supplied, this parameter
+                should be a list of path patterns e.g. ``["foo/**/*.py"]``.
```
lurch, line 103: Not sure if it makes sense for the example patterns to be the same for both `filter_glob` and `exclude_glob`? :wink: Perhaps `["foo/**/*.pyc"]` would be a more useful example for the latter?

fs/glob.py
```diff
@@ -91,9 +172,95 @@ def imatch(pattern, path):
...
+        >>> from fs import wildcard
+        >>> is_python = wildcard.get_matcher(['*.py'], True)
+        >>> is_python('__init__.py')
+        True
+        >>> is_python('foo.txt')
+        False
+
+    """
+    if not patterns:
+        return lambda name: True
```
lurch, line 247: Doesn't really matter, but perhaps `name` here should be `path`? (to match up with the `path` parameter-name in `match_any` and `imatch_any`)

fs/glob.py
```diff
@@ -91,9 +172,95 @@ def imatch(pattern, path):
...
+
+    """
+    if not patterns:
+        return True
+    return any(imatch(pattern, path) for pattern in patterns)
+
+
+def get_matcher(patterns, case_sensitive, accept_prefix=False):
+    # type: (Iterable[Text], bool, bool) -> Callable[[Text], bool]
+    """Get a callable that matches paths against the given patterns.
```
lurch, line 222: Perhaps this should also have (a suitable variant of) the "Will return `True` if ``patterns`` is an empty list." text?

fs/glob.py
```diff
@@ -91,9 +172,95 @@ def imatch(pattern, path):
...
+            not required to match the wildcards themselves
+            but only need to be a prefix of a string that does.
+
+    Returns:
+        callable: a matcher that will return `True` if the paths given as
+        an argument matches any of the given patterns.
+
+    Example:
+        >>> from fs import wildcard
+        >>> is_python = wildcard.get_matcher(['*.py'], True)
```
lurch, line 239: Is this supposed to be `glob` rather than `wildcard` ?

fs/walk.py
```diff
@@ -85,11 +87,22 @@ def __init__(
                 any of these patterns will be removed from the walk.
             filter_dirs (list, optional): A list of patterns that will be used
                 to match directories paths. The walk will only open directories
```
lurch, line 89: Would it make sense to change "directories paths" here to "directory names"? :shrug:

fs/walk.py
```diff
@@ -198,10 +215,19 @@ def _iter_walk(
     def _check_open_dir(self, fs, path, info):
         # type: (FS, Text, Info) -> bool
         """Check if a directory should be considered in the walk."""
+        full_path = ("" if path == "/" else path) + "/" + info.name
```
lurch, line 218: Is this a clunky way of doing https://docs.pyfilesystem.org/en/latest/reference/path.html#fs.path.combine ?

fs/glob.py
```diff
@@ -27,17 +36,87 @@
...
+    return [pattern[i + 1 : j] for i, j in zip(indices[:-1], indices[1:])]
+
+
+def _translate(pattern, case_sensitive=True):
+    # type: (Text, bool) -> Text
+    """Translate a wildcard pattern to a regular expression.
+
+    There is no way to quote meta-characters.
+    Arguments:
+        pattern (str): A wildcard pattern.
```
lurch, line 65: A glob pattern?

fs/glob.py
atollk: Only with the current implementation of `abspath`, not with the general idea.

fs/base.py
atollk: The summaries (first sentences) differ: `Check if a name matches any of a list of wildcards.` and `Check if a path matches any of a list of glob patterns.`.

fs/glob.py
atollk: As the function summary says: "Compare a glob pattern with a path (case sensitive)." So `True` is correct.

wayconA:
Please promote this PR I'm desperately in need of that feature. 
@atollk 
@lurch 
@althonos

althonos:
@wayconA

Sorry for not getting this merged sooner, I've been under lots of work for my research projects. 

@atollk has definitely done a lot of work on this, and the reason I've been stalling this PR a bit is probably because I was not 100% happy with the API, but honestly I can't come up with something better, so I will merge this ASAP, and we can always refactor in the `pyfilesystem3` release (which I swear is not vaporwave).

Wallaku:
What needs to be done for this PR to be approved?
I'm determined to help.
@atollk 
@lurch 
@althonos

atollk:
done

Wallaku:
GREAT!!! Thank you very much @atollk 

Shall it be merged now?
@althonos @lurch

althonos (approved):

althonos:
Great job @atollk, and thanks for the review @lurch.

Wallaku:
Great! When the new release will be released? @althonos @lurch

althonos:
I'd just like #542 to be merged first and then we'll get a new release :+1:

PR #479, opened 2021-04-27
Some setinfo fixes

atollk:
## Type of changes

- Bug fix

## Checklist

## Description
Unit tests for `FS.setinfo` now function correctly, whereas they basically did nothing before. Also, a bug in OSFS was fixed which caused timestamps to lose precision with `setinfo`.


althonos:
Good catches, thanks.

Issue #486, opened 2021-06-18
MountFS's listdir doesn't take into account existing mounts


Frankkkkk:
Hi,
First of all thank you for this really good lib !

I'm hitting a small bug using MountFS: the `listdir` method doesn't take into account mounts for dir listing.

here's a simple test case:
```python3
#!/usr/bin/env python3
from fs.osfs import OSFS
from fs.mountfs import MountFS


mounted = OSFS('dir_a')
original = OSFS('dir_b')

final = MountFS()
final.mount('the_mount', mounted)
final.mount('', original)


print(f'Final FS /: {final.listdir("/")}')
print(f'mounted FS /: {mounted.listdir("/")}')
print(f'Final FS /the_mount: {final.listdir("/the_mount")}')
```

My fs is:
```bash
$ tree dir_a dir_b
dir_a
└── file_a
dir_b
└── file_b
```

and thus, the resulting `final` FS looks like so:
```bash
.
├── file_a
└── the_mount
    └── file_b
```

However running the script results in:
```
Final FS /: ['file_b']
mounted FS /: ['file_a']
Final FS /the_mount: ['file_a']
```

As we can see, the `mounted` FS is not taken into account by `listdir`, however it exists in the `MountFS` as show by the two last lines. The expected result should be:
```
Final FS /: ['file_b', 'the_mount']
```

I could try to hack something up but it would entail adding a final list to https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/mountfs.py#L159 , what do you think ?

Thanks  and cheers !
Frank

lurch:
From a quick look at the `mountfs.py` file, it looks like things might be "going wrong" because of your usage of `''` as the mount-path in `final.mount('', original)`.
However I dunno if the "bug" is that this behaviour should be allowed by mountfs, and mountfs is behaving incorrectly; or if mountfs should instead prevent empty strings being used as the mount-path? That's probably a decision for @willmcgugan to make...

Frankkkkk:
Indeed, that's a really good point.

However, _if_ we prevent empty strings, I see no other alternative of doing the thing above, thus we would be reducing the featureset of pyfilesystem, which would be a shame ? I'm happy to propose a MR next week if that's okay with you.

Cheers and have a good WK !

lurch:
[...]

PR #487, opened 2021-06-20
MountFS fixes

lurch:
## Type of changes

- Bug fix
- Tests

## Checklist

## Description

Addresses issues outlined #486 
* Don't allow MountFS to have an empty mount-path
* Don't allow files/dirs to be created inside MountFS's internal MemoryFS
* Change MountFS's .mounts attribute to a dict instead of a list of tuples
* Add a MountFS.unmount method
* Add associated unit-tests

It's a long time since I've done any PyFilesystem development, so this should should be considered "WIP code", comments welcome...  (e.g. I'm not sure if the `unmount` method is actually useful?)

This should be reviewed with care as the "bugfixes" this includes (the first two bullet-points above) subtly change the API of MountFS.

### DIFF
--- a/fs/mountfs.py
+++ b/fs/mountfs.py
@@ -13,8 +13,10 @@ from . import errors
 from .base import FS
 from .memoryfs import MemoryFS
 from .path import abspath
+from .path import dirname
 from .path import forcedir
 from .path import normpath
+from .path import recursepath
 from .mode import validate_open_mode
 from .mode import validate_openbin_mode
 
@@ -23,10 +25,10 @@ if typing.TYPE_CHECKING:
         Any,
         BinaryIO,
         Collection,
+        Dict,
         Iterator,
         IO,
         List,
-        MutableSequence,
         Optional,
         Text,
         Tuple,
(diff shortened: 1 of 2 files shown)


fs/mountfs.py
```diff
@@ -67,7 +69,7 @@ def __init__(self, auto_close=True):
         super(MountFS, self).__init__()
         self.auto_close = auto_close
         self.default_fs = MemoryFS()  # type: FS
-        self.mounts = []  # type: MutableSequence[Tuple[Text, FS]]
+        self.mounts = {}  # type: Dict[Text, FS]
```
althonos, line 72: I think the reason a `list` was in use here was to preserve ordering of mounted files. Sure, a `dict` is fine for newer Python versions but I'd think this would change behaviour on Python 2.7. I think using `collections.OrderedDict` would work instead (and we can drop that when we move the code to PyFilesystem3).

althonos (changes requested):
Hi @lurch , thanks a lot for this, sorry it went under my radar! I think it looks fine apart from the one change I requested (it's probably me being paranoid/thinking too hard about edge cases).

Issue #238, opened 2018-12-08
Question: pathlib compatibility?


smheidrich:
Are there any plans to make the functionality offered by this package available through an interface mirroring the Python standard library's [``pathlib.Path``](https://docs.python.org/library/pathlib.html), or does anyone know of external projects working towards this? I think adoption of ``pathlib`` is already comparatively widespread, so if something like this existed, a decent number of projects would obtain support for the various filesystems offered by PyFilesystem with no or minimal changes.

smheidrich:
The way I see it, implementing the ``PathLike`` interface (``__fspath__``) would be pretty pointless, because its only purpose is to return the path as a string so that it can be used in the parts of the standard library which are mostly just wrappers around syscalls (``open``, ``os.mkdir``, ...). This wouldn't be useful when dealing with any of the interesting filesystems supported by PyFilesystem (remote, in-memory, ...).

But what I meant was the ``pathlib.Path`` interface, which defines methods for (almost) anything one needs when dealing with files, such as [``Path.open``](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open), [``Path.iterdir``](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir), and all the methods in the right column [here](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module). Not all of these methods make sense for all of the filesystems supported by PyFilesystem, but I'd imagine the most commonly used ones are things like ``open``, ``iterdir``, ``exists``, ``mkdir``, ``rename``, ... which map nicely onto [``fs.base.open``](https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.open), [``fs.base.listdir``](https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.listdir), [``fs.base.exists``](https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.exists), [``fs.base.makedir``](https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.makedir), [``fs.base.move``](https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.move), ... you get the idea.

[...]

PR #488, opened 2021-06-23
errors: make FSError inherit from OSError

Frankkkkk:
## Type of changes

- Bug fix

## Checklist

- [😄] I accept that @PyFilesystem/maintainers may be pedantic in the code review.

## Description
Hi,

`FSError` has the same semantics than python's base `OSError`. By making
it inherit from it we enable a better transparency between python "raw"
FS methods and PyFS ones (no need to catch both).

Signed-off-by: Frank Villaro-Dixon <frank.villaro@infomaniak.com>

### DIFF
--- a/fs/errors.py
+++ b/fs/errors.py
@@ -68,7 +68,7 @@ class MissingInfoNamespace(AttributeError):
 
 
 @six.python_2_unicode_compatible
-class FSError(Exception):
+class FSError(OSError):
     """Base exception for the `fs` module."""
 
     default_message = "Unspecified error"


willmcgugan:
👎 OSErrors and FSErrors are two quite disparate things. It would be very rare to want to catch both.

lurch:
> we enable a better transparency between python "raw" FS methods and PyFS ones

IMHO that's something to avoid. Seems a lot of "new" people to PyFS get confused between "the pyfilesystem world" and the "raw python OS world".
If you're using python "raw" FS methods (where paths can have drive-letters and optional backward-slashes on Windows), then you should catch OSError.
If you're using PyFS methods (where (absolute) paths always start with a `/` and only use forward-slashes, even on Windows), then you should catch FSError.

Frankkkkk:
The use case was replacing an FS class with a PyFS one, so that would make the change transparent. But I see the nuance though. Thanks for the comments 👍

lurch:
See also #238 for a similar discussion.

PR #489, opened 2021-07-16
Use platformdirs instead of appdirs

felixonmars:
## Type of changes

- Other

## Checklist

## Description

Use the better maintained `platformdirs` instead of `appdirs`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Changed
+
+- Replaced `appdirs` with `platformdirs` dependency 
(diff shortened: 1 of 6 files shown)


althonos:
Hi @felixonmars ,

In the current state, I'll have to reject this pull request. `platformdirs ~=2.0.2` only supports Python 3, while we try to maintain Python 2.7 compatibilty before the next major release. Maybe try to use conditional specifier for Python2.7 using an older version, if possible?

felixonmars:
> `platformdirs ~=2.0.2` only supports Python 3

This is not true. Their Python 3 only version is not yet released, 2.0.2 on PyPI is still compatible with Python 2.

althonos:
Thank you! Sorry for stalling that for so long.

PR #491, opened 2021-07-29
Fix ResourceLocked translation on Windows #484

### LINKED ISSUE #484: Bug in convert_os_errors
https://github.com/PyFilesystem/pyfilesystem2/blob/2d0ffc34a2127db3fdf0f8fef6f4644bf7500a1e/fs/error_tools.py#L87

This compares a tuple to int and is always false.

dargueta:
## Type of changes

Bug fix

## Checklist

## Description

Fixed #484


fs/error_tools.py
```diff
@@ -84,7 +84,8 @@ def __exit__(
             _errno = exc_value.errno
             fserror = os_errors.get(_errno, errors.OperationFailed)
             if _errno == errno.EACCES and sys.platform == "win32":
-                if getattr(exc_value, "args", None) == 32:  # pragma: no cover
+                error_args = getattr(exc_value, "args", (None,))
+                if error_args and error_args[0] == 32:  # pragma: no cover
```
lurch, line 88: A non-empty tuple (even `(None,)`) evaluates to `True`, so is the `error_args and` part needed in the `if`?
```python
>>> bool((None,))
True
```

tests/test_error_tools.py
```diff
@@ -23,3 +24,13 @@ def test_convert_enametoolong(self):
                 raise exception
         self.assertEqual(ctx.exception.exc, exception)
         self.assertEqual(ctx.exception.path, "/tmp/test")
+
+    @unittest.skipIf(sys.platform != "win32", "requires Windows")
+    def test_convert_resourcelocked_windows(self):
```
lurch, line 29: This is failing on AppVeyor? :confused:

tests/test_error_tools.py
```diff
@@ -23,3 +24,13 @@ def test_convert_enametoolong(self):
...
+        exception = OSError(32, "resource locked")
```
lurch, line 30: Ahh, digging into it a bit it looks like you might need to pass this `32` through as the `winerror` parameter to the `OSError` constructor? https://docs.python.org/3/library/exceptions.html#OSError

fs/error_tools.py
dargueta: I included this because `OSError` can be thrown without arguments, in which case `args` will be empty.

tests/test_error_tools.py
dargueta: Well that's... odd... Somehow that feels wrong to me.

fs/error_tools.py
lurch: I wonder if it's worth just returning a plain None as the `getattr` default case then? i.e. `error_args = getattr(exc_value, "args", None)`

dargueta: That was my first thought, but then I figured if we add MyPy there's a good chance it would get angry about that, so I used a tuple instead.

tests/test_error_tools.py
lurch: Probably because POSIX and Win32 use different error codes for different things? :shrug:

dargueta: No I meant the first argument should always be the error code, at least in theory. I don't know why if we pass the error code in as the first argument we don't get it back out as the first argument in `args`.

lurch: Hmmm, that's a good point - I wonder if it suggests that the original code was wrong, and it should have actually been `if getattr(exc_value, "winerror", None) == 32:` ? :shrug:

dargueta: From the docs:

> Under Windows, this gives you the native Windows error code. The errno attribute is then an approximate translation, in POSIX terms, of that native error code.
>
> Under Windows, if the winerror constructor argument is an integer, the errno attribute is determined from the Windows error code, and the errno argument is ignored. On other platforms, the winerror argument is ignored, and the winerror attribute does not exist.

This sounds to me like we shouldn't be touching it, and if it does exist it won't be 32.

lurch: Checking back through the history, this code was [originally added](https://github.com/PyFilesystem/pyfilesystem2/commit/b32e0ec1fbebe933af54ebe209d7fb67ef057b13) by @willmcgugan - I know it's unlikely after 5 years, but I wonder if he might remember what it was _supposed_ to be doing?

Although looking back even further it seems this code came from https://github.com/PyFilesystem/pyfilesystem/blob/master/fs/errors.py#L297 ? Which was [added in 2009](https://github.com/PyFilesystem/pyfilesystem/commit/de98e4d839784e12f5f7a015eac5cf8bdd3bd43a), oh my! :open_mouth: 
Probably needs someone to experiment with actually raising the exception on Windows, rather than trying to guess at "manually reproducing" the matching exception? I guess there's a possibility that this code may not even be needed now, given the changes to both Windows and Python over the past 12 years? :man_shrugging:

willmcgugan: Haven't a clue. I'm afraid I had to dump a lot of my long term memory to make space for new Javascript libraries.

Some of that code was deduced experimentally. I do remember seeing behaviour that did not match any docs I could find. It could very well be outdated now.

dargueta: Any other ideas?

fs/error_tools.py
isaacHuh:
```suggestion
                error_args = getattr(exc_value, "args", (None,))
                if error_args and error_args[0] == 13:  # pragma: no cover
```
should check if that argument is 13 instead

isaacHuh: Also this line https://github.com/PyFilesystem/pyfilesystem2/pull/491/files#R63 
`FILE_ERRORS[13] = errors.FileExpected `
I don't get it seems wrong since the error code 13 should be resource locked error and I don't think it would ever throw file expected in this situation anyway since fserror gets overwritten later. Actually you could probably just do:

`FILE_ERRORS[13] = errors.ResourceLocked`

and not do all that if _errno == errno.EACCES and checking stuff there instead. I think this is right?

isaacHuh: @dargueta yeah this works too it seems:
```
  if _WINDOWS_PLATFORM:  # pragma: no cover
      DIR_ERRORS[13] = errors.DirectoryExpected
      DIR_ERRORS[267] = errors.DirectoryExpected
      FILE_ERRORS[13] = errors.ResourceLocked

  def __init__(self, opname, path, directory=False):
      # type: (Text, Text, bool) -> None
      self._opname = opname
      self._path = path
      self._directory = directory
      
  def __enter__(self):
      # type: () -> _ConvertOSErrors
      return self

  def __exit__(
      self,
      exc_type,  # type: Optional[Type[BaseException]]
      exc_value,  # type: Optional[BaseException]
      traceback,  # type: Optional[TracebackType]
  ):
      # type: (...) -> None
      os_errors = self.DIR_ERRORS if self._directory else self.FILE_ERRORS
      if exc_type and isinstance(exc_value, EnvironmentError):
          _errno = exc_value.errno
          fserror = os_errors.get(_errno, errors.OperationFailed)
          reraise(fserror, fserror(self._path, exc=exc_value), traceback)
```

just might wanna look as to why this line was added:
`FILE_ERRORS[13] = errors.FileExpected`

dargueta: So here's the funny thing -- according to [this](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-)

Error 32 on Windows is `ERROR_SHARING_VIOLATION`: "The process cannot access the file because it is being used by another process." This makes the most sense for the original code of the `if` statement here.

Error 13 is `ERROR_INVALID_DATA` on Windows which might be the correct one for "file expected" but I'm not sure, given there are things like `ERROR_DIRECTORY_NOT_SUPPORTED`.

_However_

POSIX error code 13 is `EACCES`, which is what you'd expect from a failed access due to a file being locked. So it looks like we should be checking `winerror` against 32, not errno?

```py
windows_error = getattr(exc_value, "winerror", 0)
exception_args = getattr(exc_value, "args", None) or (0,)
if windows_error == 32 or exception_args[0] == errno.EACCES:
    # proper exception type is errors.ResourceLocked
```

isaacHuh: @dargueta sorry for not getting back sooner I got really busy with some stuff and forgot. Idk how much more you've looked at this since then but I found some stuff that helped me understand it better. 

I was playing around with another test that raised this:
`pywintypes.error: (32, 'CreateFileW', 'The process cannot access the file because it is being used by another process.')
`
which is the windows error you'd expect. 

Seems to me that there is a step in between where you are expected to convert this windows error to oserror yourself by doing something like:
`exception = OSError(32, "resource locked")
exception.errno = 13`
since an oserror like that won't get thrown but does seem jank since you would expect args[0] to match errno at least I think its supposed to?
 
If this is the expected use here and I think it is, would be nice to include a winerror to oserror decorator or something like it in error tools or a comment explaining it or improving this to convert the winerror directly to fserror. Or at least that's my opinion on it.

dargueta: I _think_ the `windows_error = getattr(exc_value, "winerror", 0)` bit handles it. I forgot to update the unit tests with the latest change though. 😅

PR #507, opened 2021-12-09
bug: fix parsing of FTP LIST command

### LINKED ISSUE #506: FTFS incorrectly parses results of LIST command
I'm connecting to a Linux FTP server that has usernames with a `.`. The [current regex](https://github.com/PyFilesystem/pyfilesystem2/blob/79a73679e5ce63db982a13d87feca13effc41daf/fs/_ftp_parse.py#L27) (`[\w\-]+`) doesn't allow this, and so no directories are listed (side note: would it make sense to error when none of the matchers match and the line isn't empty? It was very confusing at first).

According to the [Ubuntu manpage for adduser](https://manpages.ubuntu.com/manpages/jammy/en/man5/adduser.conf.5.html#notes):

> adduser  and addgroup enforce conformity to IEEE Std 1003.1-2001, which allows only  the following characters to appear  in  group  and  user  names:  letters,  digits, underscores,  periods,  at signs (@) and dashes. The name may no start with a dash. The "$" sign is allowed at the end of usernames (to conform to samba).

Now, some distros may have more stringent restrictions, but I suggest that the regex be expanded to at least accept the pattern mentioned in the manpage for Ubuntu. I think the regex would be `[A-Za-z0-9][A-Za-z0-9\-\.\_\@]*[\$]{0,1}`.

If you agree, I opened to address this.

adriangb:
Fixes #506


fs/_ftp_parse.py
```diff
@@ -24,9 +24,9 @@
     \s+?
     (\d+)
     \s+?
-    ([\w\-]+)
+    ([A-Za-z0-9][A-Za-z0-9\-\.\_\@]*[\$]{0,1})
```
althonos, line 27: You can replace `[\$]{0,1}` with just `\$?` here.

althonos (changes requested):

althonos:
Nice job! Could you please update the `CHANGELOG.md` and `CONTRIBUTORS.md` files as well?

fs/_ftp_parse.py
adriangb: Yep!

adriangb:
> Nice job! Could you please update the `CHANGELOG.md` and `CONTRIBUTORS.md` files as well?

Done. I ended up first in the contributors list 😅 .

adriangb:
Apologies @althonos, I had made a branch off of my branch for internal use since this one will be deleted when this gets merged, so logically I committed to the wrong branch 🙃. I just pushed the changes mentioned above.

althonos:
Perfect, thanks!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Support more leniant usernames and group names in FTP servers [#507](https://github.com/PyFilesystem/pyfilesystem2/pull/507). Closes [#506](https://github.com/PyFilesystem/pyfilesystem2/pull/506).
```
lurch, line 11: Too late for this PR, but perhaps you could fix 'leniant' -> 'lenient' in your next PR? :slightly_smiling_face:

adriangb:
Hi @althonos , is there any chance you can cut a release with this change? Thanks!

althonos:
@adriangb : done now!

Issue #515, opened 2022-01-27
Is there a way to check whether two paths (or filesystems) are referencing the same file?


tfeldmann:
I'd like to know if there is a way to check whether two pathes reference the same file.
I found that not even the filesystems can be compared:

```python
>>> import fs
>>> fs1 = fs.open_fs("~/Desktop")
>>> fs2 = fs.open_fs("~/Desktop")
>>> fs1 == fs2
False
```

At the moment I'm resorting to this, which is not nice and complete at all.

```python
from typing import NamedTuple

class Resource(NamedTuple):
    fs: FS
    path: str

def is_same(a: Resource, b: Resource):
    from fs.osfs import OSFS
    from fs.zipfs import ZipFS
    from fs.tarfs import TarFS

    if isinstance(a.fs, OSFS) and isinstance(b.fs, OSFS):
        return a.fs.getospath(a.path) == b.fs.getospath(b.path)
    elif isinstance(a.fs, ZipFS) and isinstance(b.fs, ZipFS):
        return a.fs._file == b.fs._file and a.path == b.path
    elif isinstance(a.fs, TarFS) and isinstance(b.fs, TarFS):
        return a.fs._file == b.fs._file and a.path == b.path
    return False
```

Is there a nicer way to check this?
Would you be interested in a PR that fixes the `__eq__` method for some filesystems?

tfeldmann:
Ok the use case is a bit complex. I'm currently migrating [organize](github.com/tfeldmann/organize) to pyfilesystem, which has a duplicate detection.

Now image users having a config with a list of (fs, path) tuples like this:
```python
"locations": [
  ("~/Desktop/", "/"),
  ("~", "/Desktop"),
  ("zip:///Users/test/myzip.zip", "/"),
]
```

For each tuple a Walker is created. All files emitted by the chained walkers are now checked for duplicates.
So without the `is_same_resource` check every file on this poor guy's desktop would be detected as a duplicate.

lurch:
Would users expect a file-inside-a-zipfile to be detected as a duplicate of a file-on-the-desktop? Obviously that'd need slightly more complex logic.

Just to be pedantic, your previous comment is a list of (fsurl, fspath) tuples :wink:

tfeldmann:
> Would users expect a file-inside-a-zipfile to be detected as a duplicate of a file-on-the-desktop? Obviously that'd need slightly more complex logic.

[...]

PR #519, opened 2022-01-28
Enable comparing filesystems with `==`

tfeldmann:
## Type of changes

- Other

## Checklist

## Description

This PR enables comparing filesystems with the `==` operator. Based on the discussion in #515.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -38,6 +38,7 @@ Many thanks to the following developers for contributing to this project:
 - [Silvan Spross](https://github.com/sspross)
 - [@sqwishy](https://github.com/sqwishy)
 - [Sven Schliesing](https://github.com/muffl0n)
+- [Thomas Feldmann](https://github.com/tfeldmann)
 - [Tim Gates](https://github.com/timgates42/)
 - [@tkossak](https://github.com/tkossak)
 - [Todd Levi](https://github.com/televi)
--- a/fs/base.py
+++ b/fs/base.py
@@ -130,6 +130,18 @@ class FS(object):
         """Close filesystem on exit."""
         self.close()
 
+    def __eq__(self, other):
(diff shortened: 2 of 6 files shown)


willmcgugan:
Testing the equality of an FS is not something that makes sense across all filesystems. Your default implementation won't work for the more virtual filesystems. Even with the same class and the same syspath for /, a subdirectory may reference different syspaths.

It's also not clear what equality means with a filesystem. A user may reasonably expect that the contents are the same.

I can see why you might need this, but it's not something that should be exposed via `__eq__`. I'd suggest defining a function with the logic you need. But it probably isn't something that belongs in the core lib.

tfeldmann:
> Testing the equality of an FS is not something that makes sense across all filesystems. Your default implementation won't work for the more virtual filesystems.

That's true. I see equality as "references the same resource", so for example `open_fs("~Desktop") == open_fs("~/Desktop")`. It's fine if the more virtual ones cannot be compared.

> Even with the same class and the same syspath for /, a subdirectory may reference different syspaths.

I didn't know this is possible - can you expand on that? I guess I misunderstood something.
Would the same ospath be a better indicator?

> It's also not clear what equality means with a filesystem. A user may reasonably expect that the contents are the same.

Never thought about it this way - good point

> I can see why you might need this, but it's not something that should be exposed via `__eq__`. I'd suggest defining a function with the logic you need. But it probably isn't something that belongs in the core lib.

That's fine, too! In this case feel free to close the PR 👍

lurch:
> > Even with the same class and the same syspath for /, a subdirectory may reference different syspaths.

> I didn't know this is possible - can you expand on that?

E.g. https://docs.pyfilesystem.org/en/latest/reference/multifs.html or https://docs.pyfilesystem.org/en/latest/reference/mountfs.html

Issue #317, opened 2019-07-28
Drop Python2.7 support


willmcgugan:
Python2.7 will be unsupported in a [few months](https://pythonclock.org/). We should drop Python2.7 before then.

I'm keen to do this as Python2.7 support has long been a headache. If anyone has a good reason to keep Py2.7 support, please make your case in this issue.

zopyx:
Many Python 2.7 rest for ever.
No need for further Python 2.7 compatibility.

althonos:
Fine by me, I'm not putting any particular effort for Python 2.7 support right now so I don't care about dropping support.
IMO you should make a last stable release so that all libraries can pin that version more easily in case some devs stuck in the past need to pin anything.

bennr01:
**Small suggestion:**  dropping py2 support is technically a backwards incompatible api change. When following [semantic versioning](https://semver.org/), this should be indicated by a major version bump.

`pyfilesystem2` currently has the major version `2`, which would mean that the next major version would be `3`. As most of you probably know, a python package ending with `3` usually indicates py3-only support, which would coincide with the change.

This would also allow py2-projects to specify `fs < 3.0.0` as a dependency.

On the other hand, bumping the major version of a package in a repo called `pyfilesystem2` may lead to confusion, so I am not sure whether this change should really be done in accordance with semantic versioning.

dargueta:
>  a python package ending with 3 usually indicates py3-only support

Package *name* or package *version*? I agree with the name.

> On the other hand, bumping the major version of a package in a repo called pyfilesystem2 may lead to confusion

I do agree that the `2` ending has put us in an unfortunate position, but we're going to have to bump the major version at some point. Otherwise we'll be violating the principle of semantic versioning.

willmcgugan:
@bennr01 I concur 3.0.0 is the way to go when we drop Python2 support.

The name on Pypy is still `fs`. I hope the repos name containing a 2 doesn't confuse too many people. I can always put a note in the readme if it does.

odgalvin:
@althonos I could go through and remove the Python 2 compatibility bits if you'd like a PR? Maybe after your next point release. It would tidy up the code a lot.

althonos:
@odgalvin : as Will mentioned, we should make a new repository for that, probably forking this one into PyFilesystem/pyfilesystem3. Plus, it would be great to take advantage of this breaking change to actually address a lot of the breaking issues.

atollk:
[...]

PR #523, opened 2022-02-08
Optimize `move.move_file` for moving files between different OSFS instances.

tfeldmann:
## Type of changes

- New feature

## Checklist

## Description

Using `move.move_file` to move a file between different OSFS instances is extremely slow because it first copies the file to the destination and then deletes the source.
This PR adds an optimization for this use case.

### DIFF
--- a/fs/move.py
+++ b/fs/move.py
@@ -4,11 +4,15 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from os.path import commonpath
 import typing
 
+from . import open_fs, errors
 from .copy import copy_dir
 from .copy import copy_file
 from .opener import manage_fs
+from .osfs import OSFS
+from .path import frombase
 
 if typing.TYPE_CHECKING:
     from .base import FS
@@ -55,6 +59,21 @@ def move_file(
             resources (defaults to `False`).
 
     """
+    # optimization for moving files between different OSFS instances
+    if isinstance(src_fs, OSFS) and isinstance(dst_fs, OSFS):
+        try:
+            src_syspath = src_fs.getsyspath(src_path)
+            dst_syspath = dst_fs.getsyspath(dst_path)
+            common = commonpath([src_syspath, dst_syspath])
+            rel_src = frombase(common, src_syspath)
+            rel_dst = frombase(common, dst_syspath)
+            with open_fs(common, writeable=True) as base:
+                base.move(rel_src, rel_dst, preserve_time=preserve_time)
+            return
+        except (ValueError, errors.NoSysPath):
+            # optimization cannot be applied
+            pass
+
     with manage_fs(src_fs) as _src_fs:
         with manage_fs(dst_fs, create=True) as _dst_fs:
             if _src_fs is _dst_fs:
--- a/tests/test_move.py
+++ b/tests/test_move.py
@@ -6,6 +6,7 @@ from parameterized import parameterized_class
 
 import fs.move
 from fs import open_fs
+from fs.path import join
 
 
 @parameterized_class(("preserve_time",), [(True,), (False,)])
@@ -36,6 +37,17 @@ class TestMove(unittest.TestCase):
             dst_file2_info = dst_fs.getinfo("foo/bar/baz.txt", namespaces)
             self.assertEqual(dst_file1_info.modified, src_file1_info.modified)
             self.assertEqual(dst_file2_info.modified, src_file2_info.modified)
+
+    def test_move_file(self):
+        with open_fs("temp://") as temp:
+            syspath = temp.getsyspath("/")
+            a = open_fs(syspath)
+            a.makedir("dir")
+            b = open_fs(join(syspath, "dir"))
+            b.writetext("file.txt", "Content")
+            fs.move.move_file(b, "file.txt", a, "here.txt")
+            self.assertEqual(a.readtext("here.txt"), "Content")
+            self.assertFalse(b.exists("file.txt"))
 
     def test_move_dir(self):
         namespaces = ("details", "modified")
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -38,6 +38,7 @@ Many thanks to the following developers for contributing to this project:
 - [Silvan Spross](https://github.com/sspross)
 - [@sqwishy](https://github.com/sqwishy)
 - [Sven Schliesing](https://github.com/muffl0n)
+- [Thomas Feldmann](https://github.com/tfeldmann)
 - [Tim Gates](https://github.com/timgates42/)
 - [@tkossak](https://github.com/tkossak)
 - [Todd Levi](https://github.com/televi)


fs/move.py
```diff
@@ -55,6 +59,21 @@ def move_file(
...
+        try:
+            src_syspath = src_fs.getsyspath(src_path)
+            dst_syspath = dst_fs.getsyspath(dst_path)
+            common = commonpath([src_syspath, dst_syspath])
+            rel_src = frombase(common, src_syspath)
+            rel_dst = frombase(common, dst_syspath)
+            with open_fs(common, writeable=True) as base:
+                base.move(rel_src, rel_dst, preserve_time=preserve_time)
+            return
+        except (ValueError, errors.NoSysPath):
```
lurch, line 73: If we know both FSes are OSFS instances, is it even possible to get a NoSysPath error?

Also, do we need to bother with calculating a commonpath, or can we just do `os.rename(src_syspath, dst_syspath)` ?


Hmmm, thinking about it some more, we could probably just do this if both `src_path` and `dst_path` have a SysPath, _regardless_ of what class their FSes are? And it probably makes sense to use the same logic that https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/base.py#L1134 does? (i.e. try to rename first, and then drop through to default behaviour). So perhaps this "optimisation" should go just above the `# Standard copy and delete` below?

Although no doubt there's intricacies that I've probably overlooked :laughing:

tfeldmann: >  we could probably just do this if both src_path and dst_path have a SysPath, regardless of what class their FSes are?

I tried that and had trouble with ZipFS

lurch: Seems to me like a ZipFS shouldn't even have a SysPath? Perhaps that's a bug in ZipFS? :man_shrugging:

althonos: Yes, in theory `ZipFS` shouldn't have a syspath, so that's a bug. But I concur with @lurch I think the easiest would be to write the code in the context (otherwise you lose the ability to optimize a copy between FS URLs) and simply do:
```python
if _src_fs is _dst_fs:
   ...
elif _src_fs.hassyspath(src_path) and _dst_fs.hassyspath(dst_path):
   ...
else:
   ...
```

tfeldmann:
@althonos Implement it as you suggested.
I guess my problems with ZipFS came from not closing the filesystem because it was not in the managed context.
The question is: Do I need to `.lock()` the filesystems?

tfeldmann:
I guess I need to lock. Otherwise the file could be deleted from src_fs during the move operation.

althonos:
I guess you just need to patch the `WrapReadOnly` methods now!

tfeldmann:
[Comment to trigger the test workflow?]

fs/move.py
tfeldmann: @lurch I think `os.rename` wouldn't work across drives on windows, but `shutil.move()` is an option. I decided against that because that would need some extra code for `preserve_time`...

tfeldmann:
Should be fine now, can I trigger the test workflow somehow?

tests/test_move.py
```diff
@@ -37,6 +38,17 @@ def test_move_fs(self):
             self.assertEqual(dst_file1_info.modified, src_file1_info.modified)
             self.assertEqual(dst_file2_info.modified, src_file2_info.modified)
 
+    def test_move_file(self):
```
lurch, line 41: It looks like this method was written explicitly to test your "optimised" code-path?
Might also be worth adding a version that moves a file from one memfs to another memfs, to test the non-optimised code-path?

fs/move.py
```diff
@@ -55,6 +59,31 @@ def move_file(
...
+        # we have to raise a ResourceReadOnly exception manually if a FS is read-only
+        if src_fs.getmeta().get("read_only", True):
+            raise ResourceReadOnly(src_fs, src_path)
+        if dst_fs.getmeta().get("read_only", True):
+            raise ResourceReadOnly(dst_fs, dst_path)
+
+        # if both filesystems have a syspath we create a new OSFS from a
+        # common parent folder and use it to move the file.
+        with manage_fs(src_fs) as _src_fs:
+            with manage_fs(dst_fs, create=True) as _dst_fs:
```
lurch, line 72: Should these `manage_fs` calls (and also the ones elsewhere in this function) all be setting `writeable=True` ? (since if we're moving a file, both the `src_fs` and `dst_fs` will end up getting modified).

tests/test_move.py
```diff
@@ -37,6 +38,17 @@ def test_move_fs(self):
...
+        with open_fs("temp://") as temp:
+            syspath = temp.getsyspath("/")
+            a = open_fs(syspath)
+            a.makedir("dir")
+            b = open_fs(join(syspath, "dir"))
+            b.writetext("file.txt", "Content")
+            fs.move.move_file(b, "file.txt", a, "here.txt")
```
lurch, line 48: Since src_fs and dst_fs can be FS objects or FS-URLs, perhaps it's worth testing passing the FSes by FS-URL also? :shrug:  (or perhaps that's unnecessary? @althonos ? )

fs/move.py
lurch: @althonos On a related note, why does the `move_dir` function elsewhere in this file use `manage_fs(src_fs, writeable=False)` ? Does that mean that theoretically the `_src_fs.removetree(src_path)` at the end of that function shouldn't work?

fs/move.py
```diff
@@ -55,6 +59,31 @@ def move_file(
             resources (defaults to `False`).
 
     """
+    if src_fs.hassyspath(src_path) and dst_fs.hassyspath(dst_path):
```
lurch, line 62: Ah! It's just occurred to me that this probably won't work if src_fs or dst_fs are strings? :stuck_out_tongue_winking_eye:  (i.e. FS-URLs)

tests/test_move.py
tfeldmann: Just added that 👍

tests/test_move.py
tfeldmann: I missed that - I'll add a test for that.

fs/move.py
tfeldmann: Good catch! On it.

fs/move.py
```diff
@@ -62,17 +66,42 @@ def move_file(
...
+                    common = commonpath([src_syspath, dst_syspath])
+                    rel_src = frombase(common, src_syspath)
+                    rel_dst = frombase(common, dst_syspath)
+                    with _src_fs.lock(), _dst_fs.lock(), open_fs(
+                        common, writeable=True
+                    ) as base:
+                        base.move(rel_src, rel_dst, preserve_time=preserve_time)
+                    return  # optimization worked, exit early
+                except ValueError:
+                    pass
```
lurch, line 93: Urgh, isn't this generally regarded as a code-smell?

fs/move.py
```diff
@@ -62,17 +66,42 @@ def move_file(
...
-                        _src_fs,
-                        src_path,
-                        _dst_fs,
-                        dst_path,
-                        preserve_time=preserve_time,
-                    )
-                    _src_fs.remove(src_path)
+                return
+
+            if _src_fs.hassyspath(src_path) and _dst_fs.hassyspath(dst_path):
```
lurch, line 71: Would
```python
if thing1:
    do1
    return

if thing2:
    do2
```
be better as:
```python
if thing1:
    do1
elif thing2:
    do2
```

?

tests/test_move.py
```diff
@@ -37,6 +38,42 @@ def test_move_fs(self):
...
+            self.assertEqual(dst.readtext("dest.txt"), "Source")
+
+    def test_move_file_fs_urls(self):
+        # create a temp dir to work on
+        with open_fs("temp://") as tmp:
+            path = tmp.getsyspath("/")
+
+            tmp.writetext("file.txt", "Content")
+            tmp.makedir("subdir")
+            fs.move.move_file(path, "file.txt", join(path, "subdir"), "file.txt")
```
lurch, line 66: I think you need to pass in fs-urls, not syspaths? See https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.geturl

fs/move.py
tfeldmann: This works for the first case, but the second one can pass or return and we need to fall through to the standard method. It comes down to whether you like early exits 🤷‍♂️

fs/move.py
tfeldmann: Then I'd need two more ifs (one for each `frombase` call) which would result in more nesting, I thought it was cleaner this way

fs/move.py
lurch: I'll leave that up to @althonos as he's much more familiar with the general codebase here than I am.

tests/test_move.py
tfeldmann: Cleared that up 👍

fs/move.py
tfeldmann: Yes that's a bug. It's possible to have a file copied over but the src is not removed. I just submitted a test case for that and try to fix this.

fs/move.py
lurch: At the very least you should add a comment in the code explaining why you're doing this?

fs/move.py
tfeldmann: Just settings these to `writeable=True` in `manage_fs` isn't enough. I'm unsure, is this a bug? Should I just delete the copy when deleting the source fails?

fs/move.py
tfeldmann: 👍 Did that

fs/move.py
tfeldmann: I submitted a proposal which cleans up correctly after itself.

fs/move.py
```diff
@@ -55,24 +59,57 @@ def move_file(
...
+                    _dst_fs,
+                    dst_path,
+                    preserve_time=preserve_time,
+                )
+                try:
                     _src_fs.remove(src_path)
+                except FSError as e:
+                    # if the source cannot be removed we delete the copy on the
+                    # destination
+                    _dst_fs.remove(dst_path)
```
lurch, line 111: Perhaps the better "fix" here is to just move your `.getmeta().get("read_only", True)` calls to be directly below the `manage_fs` lines?

fs/move.py
lurch: @althonos On _another_ sidenote, perhaps https://github.com/PyFilesystem/pyfilesystem2/blob/f6a6195d10452da79b0068a7003042fa03c87ab4/fs/opener/registry.py#L285-L286 should check that the FS-object is writable if `writeable=True` is passed in, which I think would also have the side-benefit of tidying up the code in this PR?

fs/move.py
tfeldmann: Can I depend on the fs being writeable if the meta info says so? What about read-only OSFS like a locked SD card or read-only linux mount?

lurch: ...or perhaps an "even better" fix is to tweak the behaviour of `manage_fs`, as I just suggested in another comment :wink:

fs/move.py
tfeldmann: Yes, I kind of expected it to do this. 👍

tfeldmann: I implemented this and cleaned up the PR.

tfeldmann:
Two questions remain:
- What is wrong with the Appveyor build not finding `six`
-  Should move_file and move_dir try to clean up the destination if the transfer is interrupted / fails?

fs/move.py
lurch: > Can I depend on the fs being writeable if the meta info says so? What about read-only OSFS like a locked SD card or read-only linux mount?

Good point - perhaps we need to add an extra keyword-arg so that the user can choose the desired behaviour if it turns out not to be possible to remove the original source copy?

tfeldmann: It's still the question whether the destination should be cleaned up in case of errors

fs/move.py
```diff
@@ -55,24 +59,50 @@ def move_file(
...
+                    src_syspath = _src_fs.getsyspath(src_path)
+                    dst_syspath = _dst_fs.getsyspath(dst_path)
+                    common = commonpath([src_syspath, dst_syspath])
+                    rel_src = frombase(common, src_syspath)
+                    rel_dst = frombase(common, dst_syspath)
+                    with _src_fs.lock(), _dst_fs.lock(), open_fs(
+                        common, writeable=True
+                    ) as base:
+                        base.move(rel_src, rel_dst, preserve_time=preserve_time)
+                    return  # optimization worked, exit early
```
lurch, line 84: Perhaps it makes sense (stylistically) to move this to the same level of indentation as `base.move` ?

fs/move.py
```diff
@@ -55,24 +59,50 @@ def move_file(
...
+                except ValueError:
+                    # This is raised if we cannot find a common base folder.
+                    # In this case just fall through to the standard method.
+                    pass
```
lurch, line 88: :+1:

fs/opener/registry.py
```diff
@@ -282,10 +283,18 @@ def manage_fs(
         """
         from ..base import FS
 
+        def assert_writeable(fs):
+            if fs.getmeta().get("read_only", True):
+                raise ResourceReadOnly(path=str(fs_url))
```
lurch, line 288: As it's the FS itself that is read_only, I wonder if the appropriate `path` to pass would be `/` rather than the `fs_url` ?

tests/test_move.py
```diff
@@ -6,10 +6,13 @@
...
 import fs.move
 from fs import open_fs
+from fs.errors import ResourceReadOnly
+from fs.path import join
+from fs.wrap import read_only
 
 
 @parameterized_class(("preserve_time",), [(True,), (False,)])
-class TestMove(unittest.TestCase):
+class TestMovePreserveTime(unittest.TestCase):
```
lurch, line 15: Perhaps this should be called `TestMoveCheckTime`, to reflect the fact that the times _aren't_ always preserved? :shrug:

tfeldmann: Yes, that's better!

tests/test_move.py
```diff
@@ -60,3 +63,91 @@ def test_move_dir(self):
...
+            syspath = temp.getsyspath("/")
+            a = open_fs(syspath)
+            a.makedir("dir")
+            b = open_fs(join(syspath, "dir"))
+            b.writetext("file.txt", "Content")
+            fs.move.move_file(b, "file.txt", a, "here.txt")
+            self.assertEqual(a.readtext("here.txt"), "Content")
+            self.assertFalse(b.exists("file.txt"))
+
+    def test_move_file_different_mems(self):
```
lurch, line 80: Feels like this ought to be renamed to `test_move_file`, modified to also check the `preserve_time`, and then moved to the `TestMovePreserveTime` class? (i.e. have it as the "base test")
And then rename `TestMove.test_move_file` to e.g. TestMove.test_move_file_optimised` or TestMove.test_move_file_tempfs` ? (since I guess you can't actually unit-test whether the optimised code-path is being followed).

fs/opener/registry.py
tfeldmann: This way results in the nicer error message:

```
fs.errors.ResourceReadOnly: resource '<osfs '/Users/tf/Desktop'>(read-only)' is read only
```

tests/test_move.py
```diff
@@ -60,3 +63,91 @@ def test_move_dir(self):
...
             self.assertEqual(dst_file2_info.modified, src_file2_info.modified)
+
+
+class TestMove(unittest.TestCase):
+    def test_move_file(self):
+        with open_fs("temp://") as temp:
+            syspath = temp.getsyspath("/")
+            a = open_fs(syspath)
+            a.makedir("dir")
+            b = open_fs(join(syspath, "dir"))
```
lurch, line 74: IMHO this looks a bit weird (creating a file in the subdirectory of a syspath of an already-existing FS).
I think your optimised code-path would still get followed if `a` and `b` were both independent TempFS instances?

tests/test_move.py
```diff
@@ -60,3 +63,91 @@ def test_move_dir(self):
...
+            )
+
+            self.assertFalse(tmp.exists("file.txt"))
+            self.assertEqual(tmp.readtext("subdir/file.txt"), "Content")
+
+        with open_fs("mem://") as src, open_fs("mem://") as dst:
+            src.writetext("source.txt", "Source")
+            fs.move.move_file(src, "source.txt", dst, "dest.txt")
+            self.assertFalse(src.exists("source.txt"))
+            self.assertEqual(dst.readtext("dest.txt"), "Source")
```
lurch, line 108: This seems to be a duplicate of the code in `test_move_file_different_mems` ??

tests/test_move.py
```diff
@@ -60,3 +63,91 @@ def test_move_dir(self):
...
+            src = read_only(open_fs(path))
+            dst = tmp.makedir("sub")
+            with self.assertRaises(ResourceReadOnly):
+                fs.move.move_file(src, "file.txt", dst, "file.txt")
+            self.assertFalse(
+                dst.exists("file.txt"), "file should not have been copied over"
+            )
+            self.assertTrue(src.exists("file.txt"))
+
+    def test_move_file_read_only_mem_source(self):
```
lurch, line 123: I guess there's no harm in also adding `test_move_file_read_only_mem_dest` ? :wink:

fs/opener/registry.py
lurch: I wasn't sure if people catching the error from `ResourceReadOnly` might always expect `path` to be a valid pyfs-path? :shrug: 
(I guess _theoretically_ there could be a different `FilesystemReadOnly` error, if we don't know which particular resource was being accessed)

fs/move.py
tfeldmann: Beautified this a bit ;)

lurch: Nice, `with a(), b(), c() as d:` _was_ a bit confusing to read :wink: 
I still think that the `return` ought to be at the same indent-level as the `base.move` call though?

tfeldmann: Sure, why not - comes in the next commit

fs/move.py
```diff
@@ -4,11 +4,15 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from os.path import commonpath
```
lurch, line 7: Unfortunately this was only added in Python 3.5 https://docs.python.org/3/library/os.path.html#os.path.commonpath so this is causing the CI to fail on Python2.7 (see also #317 )

tfeldmann: Would it be ok if I copy this over to fs.path?

lurch: (perhaps it would make a useful addition to https://docs.pyfilesystem.org/en/latest/reference/path.html ? Seems like it's be fairly to implement using `fs.path.recursepath`?)

tfeldmann: Urgh, 2.7 is kind of a bummer

tfeldmann: Ok so I'm planning to add `_pathcompat.py` and backport the function there

fs/opener/registry.py
tfeldmann: Ok, better to be consistent - changed it 👍

lurch:
> What is wrong with the Appveyor build not finding six

Looking through the Appveyor history, it seems that https://ci.appveyor.com/project/willmcgugan/pyfilesystem2/builds/43007961 worked for both Py3.6 and Py3.7, but https://ci.appveyor.com/project/willmcgugan/pyfilesystem2/builds/43015535 worked for Py3.6 but failed for Py3.7, so maybe it's just an Appveyor problem? :man_shrugging:

fs/move.py
```diff
@@ -1,19 +1,22 @@
...
-from __future__ import print_function
-from __future__ import unicode_literals
+from __future__ import print_function, unicode_literals
 
 import typing
 
-from .copy import copy_dir
-from .copy import copy_file
+from . import open_fs
+from ._pathcompat import commonpath
```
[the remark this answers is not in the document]
tfeldmann: You mean the sorting? Sure, sry I had isort on muscle memory.

fs/move.py
lurch: @tfeldmann :point_up: in case you missed this comment / question :wink:

tests/test_move.py
```diff
@@ -60,3 +82,91 @@ def test_move_dir(self):
         if self.preserve_time:
             dst_file2_info = dst_fs.getinfo("bar/baz.txt", namespaces)
             self.assertEqual(dst_file2_info.modified, src_file2_info.modified)
+
+
+class TestMove(unittest.TestCase):
+    def test_move_file_tempfs(self):
+        with open_fs("temp://") as a, open_fs("temp://") as b:
+            dir_a = a.makedir("dir")
+            dir_b = b.makedir("subdir")
```
lurch, line 91: Now that you've "fixed" it so that `a` and `b` are independent FSes, I think you probably don't need these subdirectories?

tests/test_move.py
```diff
@@ -60,3 +82,91 @@ def test_move_dir(self):
...
+    def test_move_file_fs_urls(self):
+        # create a temp dir to work on
+        with open_fs("temp://") as tmp:
+            path = tmp.getsyspath("/")
+            tmp.writetext("file.txt", "Content")
+            tmp.makedir("subdir")
+            fs.move.move_file(
+                "osfs://" + path,
+                "file.txt",
+                "osfs://" + join(path, "subdir"),
```
lurch, line 106: Again, I think these ought to be independent TempFSes, rather than one being in a subdirectory of the other. IMHO.

tests/test_move.py
tfeldmann: not needed, but it's nice to test the base / common logic a bit?

tests/test_move.py
lurch: You've marked this as resolved, but I don't think this chunk of code belongs in `test_move_file_fs_urls` ?

tests/test_move.py
tfeldmann: Ok, I'll look into this in the coming days, thank you for the thorough review 👍

tests/test_move.py
tfeldmann: Totally missed that!

fs/move.py
lurch: Just had a thought... we probably want the same `cleanup_dest_on_error` parameter for both `move_file` and `move_dir`?
And in the `move_dir` case, if we've moved files `a`, `b` and `c` from `src` to `dest` (and then deleted them from `src`), but we run into an error in transferring `d` (e.g. it's a network filesystem and the connection drops), we probably **don't** want to then delete `a`, `b` and `c` from `dest`?

So I guess we could either have `move_dir` defaulting `cleanup_dest_on_error` to `False`, and `move_file` defaulting it to `True`; or maybe it's more consistent to default it to `False` for both functions? :thinking: :man_shrugging:  What do the other @PyFilesystem/maintainers think?

tfeldmann: I was just in the middle of implementing this and had the same thought. Another option is to track which files were copied over and just clean up those?

tests/test_move.py
lurch: Ahh, true. But I think it would be better if the FSes didn't "overlap", e.g. you could do something like `"osfs://" + join(path, "subdir1")` for the first FS-URL and `"osfs://" + join(path, "subdir2")` for the second FS-URL?

fs/move.py
tfeldmann: Cleaning up after `move_dir` could get quite complicated. Suppose you copy into a folder already containing other files, the cleanup should in no case just use `removetree`.

tests/test_move.py
tfeldmann: Yes sound good

fs/move.py
lurch: > Suppose you copy into a folder already containing other files...

Urgh, good point - perhaps that's a good enough reason to go for "_not_ have a `cleanup_dest_on_error` parameter for `move_dir`" :slightly_smiling_face: 

Although it does prompt an interesting question - if the `dest_fs` _already_ had a `fileA`, and you did `move_file(src_fs, 'fileA', dest_fs, 'fileA')`, would deleting `fileA` from `dest_fs` still be the right thing to do if deleting `fileA` from `src_fs` failed? :thinking:  (IYSWIM)

fs/move.py
```diff
@@ -57,6 +58,8 @@ def move_file(
         dst_path (str): Path to a file on ``dst_fs``.
         preserve_time (bool): If `True`, try to preserve mtime of the
             resources (defaults to `False`).
+        cleanup_dest_on_error (bool): If `True`, tries to delete the file copied to
+            dst_fs if deleting the file from src_fs fails.
```
lurch, line 62: This docstring should probably also mention the default?

fs/move.py
```diff
@@ -53,26 +58,54 @@ def move_file(
...
+                # if both filesystems have a syspath we create a new OSFS from a
+                # common parent folder and use it to move the file.
+                try:
+                    src_syspath = _src_fs.getsyspath(src_path)
+                    dst_syspath = _dst_fs.getsyspath(dst_path)
+                    common = commonpath([src_syspath, dst_syspath])
+                    rel_src = frombase(common, src_syspath)
+                    rel_dst = frombase(common, dst_syspath)
+                    with _src_fs.lock(), _dst_fs.lock():
+                        with open_fs(common, writeable=True) as base:
```
lurch, line 84: Given that we know that `common` is a syspath, I guess we could just use the `OSFS` constructor directly here, and "bypass" the `open_fs` machinery?

fs/move.py
```diff
@@ -53,26 +58,54 @@ def move_file(
...
-                    )
+                return
+
+            if _src_fs.hassyspath(src_path) and _dst_fs.hassyspath(dst_path):
+                # if both filesystems have a syspath we create a new OSFS from a
+                # common parent folder and use it to move the file.
+                try:
+                    src_syspath = _src_fs.getsyspath(src_path)
+                    dst_syspath = _dst_fs.getsyspath(dst_path)
+                    common = commonpath([src_syspath, dst_syspath])
```
lurch, line 80: Is it worth adding a check here that `common` isn't an empty string? Which might e.g. happen on Windows with different drive-letters?

fs/move.py
```diff
@@ -99,10 +132,10 @@ def move_dir(
...
 
     def src():
-        return manage_fs(src_fs, writeable=False)
+        return manage_fs(src_fs, writeable=True)
 
     def dst():
-        return manage_fs(dst_fs, create=True)
+        return manage_fs(dst_fs, writeable=True, create=True)
 
     with src() as _src_fs, dst() as _dst_fs:
```
lurch, line 140: Perhaps the `src()` and `dst` functions here are simple enough that the functions ought to be removed, and just move the code inline; like is done in `move_file` ?

tests/test_move.py
lurch: Fair enough. Might be nice to use more helpful variable-names than `a`, `b`, `dir_a` and `dir_b` though? :wink:

tests/test_move.py
```diff
@@ -60,3 +88,88 @@ def test_move_dir(self):
...
+        with open_fs("temp://") as tmp:
+            path = tmp.getsyspath("/")
+            subdir_src = tmp.makedir("subdir_src")
+            subdir_src.writetext("file.txt", "Content")
+            tmp.makedir("subdir_dst")
+            fs.move.move_file(
+                "osfs://" + join(path, "subdir_src"),
+                "file.txt",
+                "osfs://" + join(path, "subdir_dst"),
+                "file.txt",
```
lurch, line 114: I wonder if it's worth using a different filename on both sides (i.e. not `"file.txt"` both times), just as an additional check that the filename parameters aren't getting muddled up, or the files aren't being accidentally moved-but-not-renamed?
(same comment applies to several of the other tests too)

fs/move.py
althonos: Absolutely, in particular because it's possible (really discouraged, but possible) to edit the default opener to something different by setting `fs.opener.registry.default_opener` to something else.

fs/move.py
althonos: I'm not super opinionated about early exits in general, I guess I'm fine with the early return because the second branch is actually also an early return on success. I'd have thought an `if/elif/else` to be better if every branch was expected to succeed, which is not the case here.

althonos:
> Looking through the Appveyor history, it seems that https://ci.appveyor.com/project/willmcgugan/pyfilesystem2/builds/43007961 worked for both Py3.6 and Py3.7, but https://ci.appveyor.com/project/willmcgugan/pyfilesystem2/builds/43015535 worked for Py3.6 but failed for Py3.7, so maybe it's just an Appveyor problem? man_shrugging

Since it's possible to get Windows tested in GitHub Actions as well, I'll try to update the CI to stop using AppVeyor since it's harder to maintain _just_ for Windows.

And thanks a lot @lurch for the super thorough review :+1:

fs/move.py
tfeldmann: Agreed. And it should be `cleanup_dst_on_error`, shouldn't it?

fs/move.py
tfeldmann: Agreed, looks cleaner.

tests/test_move.py
tfeldmann: Done

althonos:
I fixed the failing tests with Python 2.7 that were caused by a circular import between several modules. The `codeformat` stage fails for a reason independent to this PR so i won't be fixing it here. Otherwise this is in really good shape, thanks a lot @tfeldmann ! 

@lurch : any last comment before i merge?

tfeldmann:
Thank you for fixing things up 👍 Looks good to me.

fs/move.py
```diff
@@ -53,26 +58,55 @@ def move_file(
         dst_path (str): Path to a file on ``dst_fs``.
         preserve_time (bool): If `True`, try to preserve mtime of the
             resources (defaults to `False`).
+        cleanup_dst_on_error (bool): If `True`, tries to delete the file copied to
+            `dst_fs` if deleting the file from `src_fs` fails (defaults to `True`).
```
lurch, line 62: Looking at the help-string above for the `src_path` and `dst_path` parameters, should the `dst_fs` and `src_fs` here be in **double** backticks? (see https://docs.pyfilesystem.org/en/latest/reference/move.html#fs.move.move_file )

fs/move.py
```diff
@@ -53,26 +58,55 @@ def move_file(
...
+                    common = commonpath([src_syspath, dst_syspath])
+                    if common:
+                        rel_src = frombase(common, src_syspath)
+                        rel_dst = frombase(common, dst_syspath)
+                        with _src_fs.lock(), _dst_fs.lock():
+                            with OSFS(common) as base:
+                                base.move(rel_src, rel_dst, preserve_time=preserve_time)
+                                return  # optimization worked, exit early
+                except ValueError:
+                    # This is raised if we cannot find a common base folder.
```
lurch, line 89: Is this exception-catching still needed, now that you check `if common:` ?

tests/test_move.py
```diff
@@ -60,3 +88,88 @@ def test_move_dir(self):
...
+
+class TestMove(unittest.TestCase):
+    def test_move_file_tempfs(self):
+        with open_fs("temp://") as src, open_fs("temp://") as dst:
+            src_dir = src.makedir("Some subfolder")
+            src_dir.writetext("file.txt", "Content")
+            dst_dir = dst.makedir("dest dir")
+            fs.move.move_file(src_dir, "file.txt", dst_dir, "target.txt")
+            self.assertEqual(dst.readtext("dest dir/target.txt"), "Content")
+            self.assertFalse(src.exists("Some subfolder/file.txt"))
```
lurch, line 101: Only a minor quibble, but some of the tests check the "new state" of the src-fs first, and some check the state of the dest-fs first. The tests might be easier to read if you always checked the FSes in a consistent order?

tests/test_move.py
```diff
@@ -60,3 +88,88 @@ def test_move_dir(self):
...
+
+    def test_move_file_fs_urls(self):
+        # create a temp dir to work on
+        with open_fs("temp://") as tmp:
+            path = tmp.getsyspath("/")
+            subdir_src = tmp.makedir("subdir_src")
+            subdir_src.writetext("file.txt", "Content")
```
lurch, line 108: _Perhaps_ it's worth changing these two lines to:
```python
tmp.makedir("subdir_src")
tmp.writetext("subdir_src/file.txt", "Content")
```
and then change line 116 to
```python
self.assertFalse(tmp.exists("subdir_src/file.txt"))
```

tests/test_move.py
```diff
@@ -60,3 +88,88 @@ def test_move_dir(self):
...
+        with open_fs("mem://") as src, open_fs("mem://") as dst:
+            src.writetext("file.txt", "Content")
+            with mock.patch.object(src, "remove") as mck:
+                mck.side_effect = FSError
+                with self.assertRaises(FSError):
+                    fs.move.move_file(
+                        src, "file.txt", dst, "target.txt", cleanup_dst_on_error=False
+                    )
+            self.assertTrue(src.exists("file.txt"))
+            self.assertTrue(dst.exists("target.txt"))
```
lurch, line 175: Is it possible to combine these two tests into one, using a similar trick to the `@parameterized_class(("preserve_time",), [(True,), (False,)])` thing?

lurch:
This is looking pretty great @tfeldmann , thank you for sticking with it through all my review comments, but hopefully you agree that the extra effort was worth it :grinning: 

> I accept that @PyFilesystem/maintainers may be pedantic in the code review.

:rofl: 


Would it be **much** more effort to do a similar "optimized-codepath-if-both-FSes-have-a-syspath" for `move.move_dir` too, or would it be better to tackle that in a separate PR?  (probably the latter, given how long this PR has taken?)

P.S. Hello from a fellow :climbing_man:

fs/move.py
tfeldmann: Yes, because `frombase` might raise one

tests/test_move.py
tfeldmann: Done

tfeldmann:
> This is looking pretty great @tfeldmann , thank you for sticking with it through all my review comments, but hopefully you agree that the extra effort was worth it 😀

Yep, definitely improved the PR a lot and I learned a thing or two 👌 Really appreciated it!

> Would it be **much** more effort to do a similar "optimized-codepath-if-both-FSes-have-a-syspath" for `move.move_dir` too, or would it be better to tackle that in a separate PR? (probably the latter, given how long this PR has taken?)

I'll tackle this in a new PR

> P.S. Hello from a fellow 🧗‍♂️

Ah, nice! 👋

tests/test_move.py
```diff
@@ -136,40 +136,35 @@ def test_move_file_read_only_mem_source(self):
...
-                        src, "file.txt", dst, "target.txt", cleanup_dst_on_error=False
+                        src,
+                        "file.txt",
+                        dst,
+                        "target.txt",
+                        cleanup_dst_on_error=cleanup,
                     )
             self.assertTrue(src.exists("file.txt"))
-            self.assertTrue(dst.exists("target.txt"))
+            self.assertEqual(not dst.exists("target.txt"), cleanup)
```
lurch, line 170: Might be marginally more readable as
```python
self.assertEqual(dst.exists("target.txt"), not cleanup)
```

lurch (approved):

tfeldmann:
Hey, any updates on this?

althonos:
This is in great shape, I'll be merging! Thanks again @tfeldmann :)

PR #527, opened 2022-03-13
Update `ReadZipFS.openbin` handler to use native Python `zipfile` code

althonos:
## Type of changes

- Bug fix

## Checklist

## Description

This PR fixes two things with the `ReadZipFS` class:

- Use the native mechanism implemented in `zipfile` for seeking files inside of Zip archives, instead of emulating it by re-opening and reading files.
- Mark `ReadZipFS` as case-sensitive, since Zip archives are case-sensitive.

### DIFF
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -332,7 +332,7 @@ class ReadZipFS(FS):
     """A readable zip file."""
 
     _meta = {
-        "case_insensitive": True,
+        "case_insensitive": False,
         "network": False,
         "read_only": True,
         "supports_rename": False,
--- a/fs/zipfs.py
+++ b/fs/zipfs.py
@@ -4,6 +4,7 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import sys
 import typing
 import zipfile
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


fs/zipfs.py
```diff
@@ -51,74 +52,126 @@ def __init__(self, fs, name):  # noqa: D107
...
+                int: the new absolute position.
+
+            Raises:
+                ValueError: when ``whence`` is not known, or ``offset``
+                    is invalid.
+
+            """
+            _whence = int(whence)
+            _pos = self.tell()
+            if _whence == Seek.current or _whence == Seek.set:
```
lurch, line 161: If `_whence` is `Seek.set`, doesn't it just need to check that `offset` is not less than 0, rather than checking `_pos + offset` ? :shrug:

althonos: Hmm, good point. I think I actually didn't implement this properly but the current tests didn't seem to catch that. Thank you for picking it up!

lurch: Haha, unlucky timing that I was leaving a comment at exactly the same time you were merging the PR! :laughing:

Issue #535, opened 2022-05-02
Issue in fs.move.move_file overwrite logic in fs==2.4.16


AleksMat:
Today's release of `fs==2.4.16` introduced an code-breaking change that isn't really documented in the release notes and might be a bug.

Here is the code sample that works for `fs<=2.4.15` but doesn't work for `fs==2.4.16`:

```Python
import fs.move
from fs.osfs import OSFS


source_fs = OSFS("source-folder")
target_fs = OSFS("target-folder")


FILENAME = "data.txt"
for filesystem in [source_fs, target_fs]:
    with filesystem.open(FILENAME, "w") as fp:
        fp.write("")

fs.move.move_file(source_fs, FILENAME, target_fs, FILENAME)
```

Basically, we have 2 different `OSFS` filesystem objects, a file in both, and we want to move a version of the file from filesystem one to another. For `fs==2.4.16` this raises an overwrite error permission:

```Python
---------------------------------------------------------------------------
DestinationExists                         Traceback (most recent call last)
Input In [1], in <module>
     11     with filesystem.open(FILENAME, "w") as fp:
     12         fp.write("")
---> 14 fs.move.move_file(source_fs, FILENAME, target_fs, FILENAME)

File ~/.local/share/virtualenvs/xxx/lib/python3.8/site-packages/fs/move.py:85, in move_file(src_fs, src_path, dst_fs, dst_path, preserve_time, cleanup_dst_on_error)
     83         with _src_fs.lock(), _dst_fs.lock():
     84             with OSFS(common) as base:
---> 85                 base.move(rel_src, rel_dst, preserve_time=preserve_time)
     86                 return  # optimization worked, exit early
     87 except ValueError:
     88     # This is raised if we cannot find a common base folder.
     89     # In this case just fall through to the standard method.

File ~/.local/share/virtualenvs/xxx/lib/python3.8/site-packages/fs/base.py:1161, in FS.move(self, src_path, dst_path, overwrite, preserve_time)
   1140 """Move a file from ``src_path`` to ``dst_path``.
   1141 
   1142 Arguments:
   (...)
   1158 
   1159 """
   1160 if not overwrite and self.exists(dst_path):
-> 1161     raise errors.DestinationExists(dst_path)
   1162 if self.getinfo(src_path).is_dir:
   1163     raise errors.FileExpected(src_path)

DestinationExists: destination '/target-folder/data.txt' exists
```

[...]

PR #542, opened 2022-07-07
Fix backward incompatibility introduced in 2.4.16 (#535)

tfeldmann:
## Type of changes

- Bug fix
- Tests

## Checklist

## Description

Added `overwrite=True` to the optimized path in `fs.move.move_file`. `move_file` now behaves like in versions <2.4.16.

### DIFF
--- a/fs/move.py
+++ b/fs/move.py
@@ -82,7 +82,12 @@ def move_file(
                         rel_dst = frombase(common, dst_syspath)
                         with _src_fs.lock(), _dst_fs.lock():
                             with OSFS(common) as base:
-                                base.move(rel_src, rel_dst, preserve_time=preserve_time)
+                                base.move(
+                                    rel_src,
+                                    rel_dst,
+                                    overwrite=True,
+                                    preserve_time=preserve_time,
+                                )
                                 return  # optimization worked, exit early
                 except ValueError:
                     # This is raised if we cannot find a common base folder.
--- a/fs/move.py
+++ b/fs/move.py
@@ -10,7 +10,7 @@ from .copy import copy_dir, copy_file
 from .errors import FSError
 from .opener import manage_fs
 from .osfs import OSFS
-from .path import frombase
+from .path import frombase, normpath
 
 if typing.TYPE_CHECKING:
     from typing import Text, Union
@@ -64,7 +64,8 @@ def move_file(
     with manage_fs(src_fs, writeable=True) as _src_fs:
         with manage_fs(dst_fs, writeable=True, create=True) as _dst_fs:
             if _src_fs is _dst_fs:
-                if src_path == dst_path:
+                # Exit early if source and destination are the same file
+                if normpath(src_path) == normpath(dst_path):
                     return
 
                 # Same filesystem, may be optimized
--- a/fs/move.py
+++ b/fs/move.py
@@ -10,7 +10,7 @@ from .copy import copy_dir, copy_file
 from .errors import FSError
 from .opener import manage_fs
 from .osfs import OSFS
-from .path import frombase, normpath
+from .path import frombase
 
 if typing.TYPE_CHECKING:
     from typing import Text, Union
--- a/fs/move.py
+++ b/fs/move.py
@@ -64,6 +64,9 @@ def move_file(
     with manage_fs(src_fs, writeable=True) as _src_fs:
         with manage_fs(dst_fs, writeable=True, create=True) as _dst_fs:
             if _src_fs is _dst_fs:
+                if src_path == dst_path:
+                    return
+
                 # Same filesystem, may be optimized
                 _src_fs.move(
                     src_path, dst_path, overwrite=True, preserve_time=preserve_time
--- a/fs/move.py
+++ b/fs/move.py
@@ -64,10 +64,6 @@ def move_file(
     with manage_fs(src_fs, writeable=True) as _src_fs:
         with manage_fs(dst_fs, writeable=True, create=True) as _dst_fs:
             if _src_fs is _dst_fs:
-                # Exit early if source and destination are the same file
-                if normpath(src_path) == normpath(dst_path):
-                    return
-
                 # Same filesystem, may be optimized
                 _src_fs.move(
                     src_path, dst_path, overwrite=True, preserve_time=preserve_time
--- a/tests/test_move.py
+++ b/tests/test_move.py
@@ -151,13 +151,29 @@ class TestMove(unittest.TestCase):
                 dst_ro.exists("target.txt"), "file should not have been copied over"
             )
 
-    def test_overwrite(self):
-        with open_fs("temp://") as src, open_fs("temp://") as dst:
-            src.writetext("file.txt", "Content")
-            dst.writetext("target.txt", "Content")
+    @parameterized.expand([("temp://",), ("mem://",)])
+    def test_move_file_overwrite(self, fs_url):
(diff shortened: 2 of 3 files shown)


tests/test_move.py
```diff
@@ -151,6 +151,14 @@ def test_move_file_read_only_mem_dest(self):
                 dst_ro.exists("target.txt"), "file should not have been copied over"
             )
 
+    def test_overwrite(self):
+        with open_fs("temp://") as src, open_fs("temp://") as dst:
```
lurch, line 155: Might be worth adding a comment here that this is specifically using TempFS instead of MemFS (as the other tests do) in order to test the syspath-optimised code?

tests/test_move.py
```diff
@@ -151,6 +151,14 @@ def test_move_file_read_only_mem_dest(self):
...
+            src.writetext("file.txt", "Content")
+            dst.writetext("target.txt", "Content")
+            fs.move.move_file(src, "file.txt", dst, "target.txt")
+            self.assertFalse(src.exists("file.txt"))
+            self.assertTrue(dst.exists("target.txt"))
```
lurch, line 160: Perhaps it's worth giving `file.txt` and `target.txt` different contents, and checking that the content of `target.txt` _has_ actually changed after the `move_file` ? :shrug:

lurch: And it might be being a bit overly-pedantic, but I wonder if it's worth adding
```python
self.assertTrue(src.exists("file.txt"))
self.assertFalse(dst.exists("file.txt"))
self.assertFalse(src.exists("target.txt"))
self.assertTrue(dst.exists("target.txt"))
```
before the `move_file`, just to make it blindingly obvious that `src` and `dst` are in fact separate FSes? (as it looks like the current test might actually pass if `src` and `dst` were pointing to the same FS?)

tests/test_move.py
tfeldmann: Good suggestion, added this!

tfeldmann:
I found another case where the optimized and the regular code path behave differently:

```python
with open_fs(fs_url) as tmp:
    tmp.writetext("file.txt", "content")
    fs.move.move_file(tmp, "file.txt", tmp, "file.txt")
```

If you move a file onto itself, the optimized code path works as expected, but the regular code path happens to delete the file. This is not wanted, I guess? I added some code to fix this.

lurch:
Eeek! If attempting to move a file onto itself, I guess it makes sense to just early-exit? :thinking:

fs/move.py
```diff
@@ -64,6 +64,9 @@ def move_file(
     with manage_fs(src_fs, writeable=True) as _src_fs:
         with manage_fs(dst_fs, writeable=True, create=True) as _dst_fs:
             if _src_fs is _dst_fs:
+                if src_path == dst_path:
+                    return
```
lurch, line 68: Does this need to do the `normpath` thing first, in case one of the paths has `../` etc.?
Also, this might need to do a case-insensitive comparison on case-insensitive filesystems?

tests/test_move.py
lurch: I was suggesting adding them _beofre_ the move_file, but you've added them _after_ ?

fs/move.py
lurch: Hmm, it's outside the scope of **this** PR, but I wonder if e.g. a `are_same_path(path1, path2)` would be a useful addition to https://docs.pyfilesystem.org/en/latest/interface.html (which would handle the normalising and case-insensitivity-ness for you automatically) ? :thinking:

tfeldmann: I don't think this is enough. We'd need a function to check if two tuples `(fs, path)` point to the same resource (I had this problem before and started working on it in #519)

lurch: Nope, I was only suggesting comparing paths (within a filesystem), NOT comparing entire filesystems, or looking at the actual resources themselves. The only "link" to the filesystem is so that it can call the `validatepath` function on each of the paths and do a case-insensitive comparison if the filesystem is itself case-insensitive. Semantically, it's only an enhanced version of `==` :slightly_smiling_face:

tfeldmann: Yes that would be a good first step - I have not tested this yet, but let's say you move a file from one connection to a ftp server onto itself in another connection to the same ftp server. This would just delete the file, wouldn't it?

This is certainly out of scope for this PR, shall I open a new issue for this?

Wallaku:
What stops this PR from being resolved?
@althonos 
@tfeldmann

lurch:
I still think you ought to do `normpath` on `src_path` and `dst_path` before comparing them, in case they contain relative-paths like `foo/../bar` ?
(also, you'll need to rebase this PR to resolve the conflicts)

tfeldmann:
Yep, some things left todo on my side. Didn't have the time yet.

tests/test_move.py
tfeldmann: Ah I see. Added it now.

fs/move.py
tfeldmann: I updated this to compare the normpathes. If you have a case-insensitive filesystem this would not exit early but delegate the move to the given FS, which should work normally but at this time triggers the delete on move bug. Exiting a little bit earlier is nice, but - as you said - a `are_same_path` function would be right approach here.

lurch:
> I found this happens also if you do `some_fs.move("file.txt", "file.txt", overwrite=True)`. It deletes the file.

@tfeldmann Can you create an issue for https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/base.py#L1138 please?

lurch (approved):
:+1:

tfeldmann:
> @tfeldmann Can you create an issue for https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/base.py#L1138 please?

On it!

fs/move.py
lurch: Created #545 for tracking this.

Wallaku:
HOORAY! Good job everybody let's bump a version for this library already!
@lurch

Wallaku:
Anyone? 
@lurch 
@althonos

fs/move.py
```diff
@@ -64,6 +64,10 @@ def move_file(
     with manage_fs(src_fs, writeable=True) as _src_fs:
         with manage_fs(dst_fs, writeable=True, create=True) as _dst_fs:
             if _src_fs is _dst_fs:
+                # Exit early if source and destination are the same file
+                if normpath(src_path) == normpath(dst_path):
+                    return
```
lurch, line 69: @tfeldmann Is this part essential for _this_ PR, or is it (and the related tests) best left for #547 ?

althonos: @lurch : this should be fine if I merge #547 first, after the rebase this should not appear as a change anymore I think.

tfeldmann: > @tfeldmann Is this part essential for _this_ PR, or is it (and the related tests) best left for #547 ?

Depends, because we use `overwrite=True` not having this would introduce a bug in this PR.

tfeldmann: We should use `validatepath` here, too. I'm fixing this up right now.

tfeldmann: Update: I found this check is no longer needed because moving onto itself is now handled correctly in `base.FS.move`.

lurch (approved):

althonos (approved):

althonos:
Great job @tfeldmann !

Issue #546, opened 2022-07-21
File is deleted when moved onto itself


tfeldmann:
Working example: 
```python
from fs import open_fs

with open_fs("mem://") as currdir:
    currdir.writetext("test.txt", "Content")
    currdir.move("test.txt", "test.txt", overwrite=True)
    assert currdir.exists("test.txt")
    # -> raises AssertionError
```

This also happens in `fs.move`.
Due to the optimization in #523 this does not happen on OSFS, but applies to other filesystems with shared connections. For example when moving a file from one connection to a ftp server onto itself in another connection to the same ftp server.

lurch:
> check whether two different `(fs, path)` tuples point to the same resource

I'm not sure there's any reasonable way that that can be done consistently for different `fs` objects? :thinking: So it should be possible to fix it for `fs.base.FS.move`, but I dunno it it'll be possible to fix it for `fs.move.move_file` ?

I guess something like
```python
from fs import open_fs
from fs.move import move_file

with open_fs("mem://") as currdir:
    currdir.makedir("subdir")
    currdir.writetext("subdir/test.txt", "Content")
    with currdir.opendir("subdir") as subdir:
        move_file(currdir, "subdir/test.txt", subdir, "test.txt")
        assert subdir.exists("test.txt")
```
might also trigger this "erroneous behaviour"?
As far as the `move_file` function is concerned, I guess it can't differentiate the above from
```python
from fs import open_fs
from fs.move import move_file

with open_fs("mem://") as currdir:
    currdir.makedir("subdir")
    currdir.writetext("subdir/test.txt", "Content")
    with open_fs("mem://") as subdir:
        subdir.writetext("test.txt", "Content")
        move_file(currdir, "subdir/test.txt", subdir, "test.txt")
        assert subdir.exists("test.txt")
```
:thinking: :man_shrugging: 

Although I guess at least for the FTP case (i.e. where we can't get a syspath), you could _try_ getting the download URL https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.geturl and use that as an attempt to see if two different `(fs, path)` tuples match to the same resource?

tfeldmann:
> I'm not sure there's any reasonable way that that can be done consistently for different fs objects? 🤔 So it should be possible to fix it for fs.base.FS.move, but I dunno it it'll be possible to fix it for fs.move.move_file ?

The `move` method needs to be patched in multiple places. I'll create a draft PR for this.

[...]

PR #547, opened 2022-07-22
Bugfix: File deleted / emptied when moved / copied onto itself (#546)

tfeldmann:
## Type of changes

- Bug fix
- Tests

## Checklist

## Description

Draft PR for fixing #546.

### DIFF
--- a/fs/base.py
+++ b/fs/base.py
@@ -1161,6 +1161,9 @@ class FS(object):
             raise errors.DestinationExists(dst_path)
         if self.getinfo(src_path).is_dir:
             raise errors.FileExpected(src_path)
+        if normpath(src_path) == normpath(dst_path):
+            # early exit when moving a file onto itself
+            return
         if self.getmeta().get("supports_rename", False):
             try:
                 src_sys_path = self.getsyspath(src_path)
--- a/fs/base.py
+++ b/fs/base.py
@@ -21,11 +21,11 @@ import warnings
 from contextlib import closing
 from functools import partial, wraps
 
-from . import copy, errors, fsencode, iotools, tools, walk, wildcard, glob
+from . import copy, errors, fsencode, glob, iotools, tools, walk, wildcard
 from .copy import copy_modified_time
 from .glob import BoundGlobber
 from .mode import validate_open_mode
-from .path import abspath, join, normpath
+from .path import abspath, isbase, join, normpath
 from .time import datetime_to_epoch
 from .walk import Walker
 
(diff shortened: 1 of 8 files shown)


tfeldmann:
My git fu is not so good. Only the latest commit is meant to be in this branch. I'll try to rebase.

lurch:
Yes, it successfully fails :laughing:

fs/osfs.py
```diff
@@ -440,6 +440,9 @@ def copy(self, src_path, dst_path, overwrite=False, preserve_time=False):
                     self.getsyspath(_src_path),
                     self.getsyspath(_dst_path),
                 )
+                # exit early if we copy the file onto itself
+                if overwrite and _src_sys == _dst_sys:
+                    return
```
lurch, line 445: Should we raise an Exception, rather than just silently returning?

tfeldmann:
The github actions probably should have a timeout 😅
At least this passes locally but can be cleaned up a bit more.

lurch:
> The github actions probably should have a timeout

I'm sure nobody would mind if you set https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes to something sensible? :smiley: 

(and I suspect #520 is probably why the whole test-suite takes so long to fail?)

althonos:
@lurch : Fine by me 👍

fs/base.py
```diff
@@ -1161,6 +1175,9 @@ def move(self, src_path, dst_path, overwrite=False, preserve_time=False):
             raise errors.DestinationExists(dst_path)
         if self.getinfo(src_path).is_dir:
             raise errors.FileExpected(src_path)
+        if normpath(src_path) == normpath(dst_path):
+            # early exit when moving a file onto itself
+            return
```
althonos, line 1180: Why use `normpath` here when you started with a `self.validatepath` call for `src_path` and `dst_path` in the other functions?

tfeldmann:
Caveat: This only fixes moving / copying on the same FS instance. I'll try to tackle the problem with the same resources on different FSs in a separate PR.

althonos:
Good job again @tfeldmann , apart from my one comment this is pretty solid and helps covering some undefined behaviour :+1: Looking forward to see how the plugins (`fs.sshfs`, `fs-s3fs`, etc.) fare with this change !

fs/base.py
```diff
@@ -1157,14 +1171,19 @@ def move(self, src_path, dst_path, overwrite=False, preserve_time=False):
...
+        _src_path = self.validatepath(src_path)
+        _dst_path = self.validatepath(dst_path)
+        if not overwrite and self.exists(_dst_path):
             raise errors.DestinationExists(dst_path)
-        if self.getinfo(src_path).is_dir:
+        if self.getinfo(_src_path).is_dir:
             raise errors.FileExpected(src_path)
+        if _src_path == _dst_path:
+            # early exit when moving a file onto itself
+            return
```
lurch, line 1182: This PR has been merged now, but I'm still not convinced that doing an early-exit here, instead of raising an `IllegalDestination`, is the correct thing to do? Since trying to move a file onto itself is _probably_ an indication of an application bug? Same comment applies to `movedir` too.

althonos: Now that you point it out, I agree. In UNIX filesystems, moving a file onto itself raises an error, so I think it's fair to expect an `IllegalDestination` there as well.

PR #550, opened 2022-08-30
Move dir optimization

tfeldmann:
## Type of changes

- New feature
- Tests

## Checklist

- [] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.

## Description

- Use `os.rename` to move folders if both filesystems have a syspath
- Check for ResourceNotFound, DirectoryExpected and IllegalDestination in `fs.move.move_dir`


tfeldmann:
Once again I should rebase this for fewer commits. And it seems to have some problems on windows.

tfeldmann:
This works now, but I'm kind of unhappy with it. For example I'm missing a param `overwrite` in `move.move_dir`. Is it ok for you if I add one? It would have to default to `true` to be backwards compatible.

fs/move.py
```diff
@@ -134,9 +138,37 @@ def move_dir(
...
+            if _src_fs.hassyspath(src_path) and _dst_fs.hassyspath(dst_path):
+                src_syspath = _src_fs.getsyspath(src_path)
+                dst_syspath = _dst_fs.getsyspath(dst_path)
+                # recheck if the move operation is legal using the syspaths
+                if isbase(src_syspath, dst_syspath):
+                    raise IllegalDestination(dst_path)
+                with _src_fs.lock(), _dst_fs.lock():
+                    with convert_os_errors("move_dir", dst_path, directory=True):
+                        if _dst_fs.exists(dst_path):
+                            os.rmdir(dst_syspath)
```
lurch, line 165: Perhaps we don't need / want this `os.rmdir`?
Looking at the non-optimised "standard copy then delete" code, it looks like perhaps if `dst_path` already exists, _and_ contains existing contents, then the existing contents will be preserved after the `move_dir` call? :shrug: 
Certainly something that it's probably worth adding an extra test for :wink:

Issue #556, opened 2022-10-26
Unable to use compress.write_zip() with s3fs source


james-emerton:
I believe this ticket should be for this package, but the issue is caused by implementation details of the fs-s3fs package. Attempting to create an archive of files from s3 using `write_zip()` fails with the following error:

```
Traceback (most recent call last):
  File "/Users/james/dev/spire/mount-s3-archive/test.py", line 7, in <module>
    write_zip(source, sys.argv[2])
  File "/Users/james/dev/spire/mount-s3-archive/.venv/lib/python3.10/site-packages/fs/compress.py", line 78, in write_zip
    mt = info.modified or datetime.utcnow()
  File "/Users/james/dev/spire/mount-s3-archive/.venv/lib/python3.10/site-packages/fs/info.py", line 322, in modified
    self._require_namespace("details")
  File "/Users/james/dev/spire/mount-s3-archive/.venv/lib/python3.10/site-packages/fs/info.py", line 128, in _require_namespace
    raise MissingInfoNamespace(namespace)
fs.errors.MissingInfoNamespace: namespace 'details' is required for this attribute
```

Code: 

```python
import sys

import fs
from fs.compress import write_zip

source = fs.open_fs(sys.argv[1])
write_zip(source, sys.argv[2])
```

Executing the above code as `test.py s3://some-bucket/ test.zip` will fail with the above error while a local source directory will compress nicely.

The issue seems to be that `S3FS` directory entries don't expose a `details` namespace. It's unclear if that's correct or not, but the code in `write_zip()` seems to assume that `details` is always available (but makes allowance for `Info.modified` to be a falsy value).

PR #557, opened 2022-10-27
#556 Fallback to alternate mtime depending correctly

james-emerton:
This will check the `stat` and `details` namespaces before defaulting to the current time.

## Type of changes

- Bug fix

## Checklist

- [X] I accept that @PyFilesystem/maintainers may be pedantic in the code review.

## Description

This addresses an issues where `write_zip()` assumed that the `details` namespace was always available. The `stat` namespace will be checked first, followed by `details`, and finally the current time will be used if `st_mtime` is still `None`.

I'm willing to author a test against this, but I'm only aware of this being an issue with the s3 filesystem (and only when it encounters a directory) and I'm not certain how best to test that here.

### DIFF
--- a/fs/compress.py
+++ b/fs/compress.py
@@ -65,17 +65,15 @@ def write_zip(
                 # Python2 expects bytes filenames
                 zip_name = zip_name.encode(encoding, "replace")
 
-            if info.has_namespace("stat"):
-                # If the file has a stat namespace, get the
-                # zip time directory from the stat structure
-                st_mtime = info.get("stat", "st_mtime", None)
-                _mtime = time.localtime(st_mtime)
-                zip_time = _mtime[0:6]  # type: ZipTime
-            else:
-                # Otherwise, use the modified time from details
-                # namespace.
-                mt = info.modified or datetime.utcnow()
-                zip_time = (mt.year, mt.month, mt.day, mt.hour, mt.minute, mt.second)
+            # If the file has a stat namespace, get the
+            # zip time directory from the stat structure
+            st_mtime = info.get("stat", "st_mtime")
+            # Otherwise, use the modified time from details namespace.
+            if st_mtime is None:
+                st_mtime = info.get("details", "modified")
+
+            # If st_mtime is still None this will default to current time
+            zip_time = time.localtime(st_mtime)[:6]  # type: ZipTime
 
             # NOTE(@althonos): typeshed's `zipfile.py` on declares
             #     ZipInfo.__init__ for Python < 3 ?!


althonos:
Hi James, sorry for the delay in answering, my real-life job has been taking a lot of my time away :sweat: 

A test would be very appreciated, as to avoid any regression (in case the code is changed or refactored, it would be better not to have this issue pop up again later). For this you can probably subclass `MemoryFS` and make a derived version of `getinfo` that never returns the `details` namespace.

CONTRIBUTORS.md
```diff
@@ -20,7 +20,9 @@ Many thanks to the following developers for contributing to this project:
 - [George Macon](https://github.com/gmacon)
 - [Giampaolo Cimino](https://github.com/gpcimino)
 - [@Hoboneer](https://github.com/Hoboneer)
+- [James Emerton](https://github.com/james-emerton)
 - [Jen Hagg](https://github.com/jenhagg)
+- [Jon Hagg](https://github.com/jon-hagg)
```
lurch, line 25: The removal of `jon-hagg` was deliberate, see #554

fs/compress.py
```diff
@@ -65,17 +65,15 @@ def write_zip(
...
-                zip_time = (mt.year, mt.month, mt.day, mt.hour, mt.minute, mt.second)
+            # If the file has a stat namespace, get the
+            # zip time directory from the stat structure
+            st_mtime = info.get("stat", "st_mtime")
+            # Otherwise, use the modified time from details namespace.
+            if st_mtime is None:
+                st_mtime = info.get("details", "modified")
+
+            # If st_mtime is still None this will default to current time
+            zip_time = time.localtime(st_mtime)[:6]  # type: ZipTime
```
lurch, line 76: In the fallback-case, does it matter that it was previously using `datetime.utcnow()` but it's now using `time.localtime()`?

james-emerton: @lurch This is a *very* good question. The datetime metadata in the ZipInfo object is meant to be localtime: https://stackoverflow.com/questions/24578625/how-to-interpret-zipinfo-date-time-value

The PyFS2 docs are mum on the format of time values, but given that `os.stat()` returns local time I believe this is correct as-is.

tests/test_zipfs.py
```diff
@@ -224,3 +226,38 @@ class TestOpener(unittest.TestCase):
     def test_not_writeable(self):
         with self.assertRaises(NotWriteable):
             open_fs("zip://foo.zip", writeable=True)
+
+
+class FSWithoutMtime(MemoryFS):
```
lurch, line 231: Just being a bit picky, but perhaps this class ought to be named `FSWithoutDetailsNamespace` ?
And perhaps it ought to be a [WrapFS](https://docs.pyfilesystem.org/en/latest/reference/wrapfs.html), rather than an explicit subclass of MemoryFS? :thinking:

tests/test_zipfs.py
```diff
@@ -224,3 +226,38 @@ class TestOpener(unittest.TestCase):
...
+    '''
+    def getinfo(self, path, namespaces):
+        if namespaces is not None:
+            namespaces = set(namespaces) - {'details'}
+        return super().getinfo(path, namespaces)
+
+
+class TestZipFSMtimeFallback(unittest.TestCase):
+    def setUp(self):
+        fh, self._temp_path = tempfile.mkstemp()
```
lurch, line 242: Would it be "better" to use a [TempFS](https://docs.pyfilesystem.org/en/latest/reference/tempfs.html) here? :shrug:

tests/test_zipfs.py
```diff
@@ -224,3 +226,38 @@ class TestOpener(unittest.TestCase):
...
+
+    def test_no_mtime(self):
+        '''Fallback to current time when creating an archive of an fs that
+        doesn't support stat or details namespaces.
+        '''
+        src_fs = FSWithoutMtime()
+        with src_fs.open('test.txt', 'w') as f:
+            f.write('Hello World')
+
+        now = datetime.now()
```
lurch, line 256: Another minor niggle, but I think this `now` variable would be better named `zip_write_datetime` or something?

tests/test_zipfs.py
james-emerton: I've renamed the class, but I don't think that subclassing `WrapFS` (which *composes* filesystems) offers any advantage over subclassing `MemoryFS` in this case.

lurch: It's https://docs.pyfilesystem.org/en/latest/reference/multifs.html that _composes_ filesystems, but I agree that for this simple test-case, what you've done here is fine.

PR #576, opened 2023-08-16
Import hook

arossert:
- New feature

Imported the [importhook](https://github.com/PyFilesystem/pyfilesystem/blob/master/fs/expose/importhook.py) from the original `PyFilesystem` project.

* Code was updated to Python3 syntax
* Added typing
* Replaced imp with importlib (in some cases I could not find the proper replacement)
* Removed unneeded code


arossert:
Is this repo still maintained?

Issue #577, opened 2023-08-23
Remove usage of deprecated module pkg_resources


matrss:
Pyfilesystem2 makes use of this module which got deprecated at some point. This leads to noisy DeprecationWarning's in the test suite of a project using pyfilesystem2.

To reproduce:
```
>>> import warnings
>>> warnings.resetwarnings()
>>> import fs
/home/icg149/mambaforge/envs/fs-report/lib/python3.11/site-packages/fs/__init__.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__("pkg_resources").declare_namespace(__name__)  # type: ignore
/home/icg149/mambaforge/envs/fs-report/lib/python3.11/site-packages/fs/__init__.py:4: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('fs')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
  __import__("pkg_resources").declare_namespace(__name__)  # type: ignore
/home/icg149/mambaforge/envs/fs-report/lib/python3.11/site-packages/fs/opener/__init__.py:6: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('fs.opener')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
  __import__("pkg_resources").declare_namespace(__name__)  # type: ignore
/home/icg149/mambaforge/envs/fs-report/lib/python3.11/site-packages/pkg_resources/__init__.py:2350: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('fs')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
  declare_namespace(parent)
```

matrss:
I just now saw that there is PR #553 for this.

mxmlnkn:
Python 3.12 did remove `pkg_resources`. Pyfilesystem2 can still be installed, but importing it will fail because of the removed `pkg_resources`.

feanil:
[...]

PR #553, opened 2022-09-27
Drop usage of `pkg_resources`

edgarrmondragon:
## Type of changes

- Bug fix

## Checklist

- [ ] I've run the latest [black](https://github.com/ambv/black) with default args on new code.
- [ ] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.
- [x] I've added tests for new code.
- [x] I accept that @PyFilesystem/maintainers may be pedantic in the code review.

## Description

Drop `pkg_resources` in favor of the `importlib.metadata` API.

Closes https://github.com/PyFilesystem/pyfilesystem2/issues/356

PR #589, opened 2024-09-16
migrate entrypoints handling to standard library importlib.metadata

eli-schwartz:
We wrap a couple different versions of the stdlib interface, and fall back to pkg_resources for compatibility with existing use cases on very old versions of python.

For python 3.8 and on, we make sure to completely avoid external dependencies, even backports.

Bug: https://github.com/PyFilesystem/pyfilesystem2/issues/577

### DIFF
--- a/fs/opener/registry.py
+++ b/fs/opener/registry.py
@@ -8,7 +8,7 @@ import typing
 
 import collections
 import contextlib
-import pkg_resources
+import sys
 
 from ..errors import ResourceReadOnly
 from .base import Opener
@@ -19,6 +19,30 @@ if typing.TYPE_CHECKING:
     from typing import Callable, Dict, Iterator, List, Text, Tuple, Type, Union
 
     from ..base import FS
+
+
+if sys.version_info >= (3, 8):
+    import importlib.metadata
+
+    if sys.version_info >= (3, 10):
+
+        def entrypoints(group, name=None):
+            ep = importlib.metadata.entry_points(group=group, name=name)
+            return tuple(n for n in ep)
+
+    else:
+
+        def entrypoints(group, name=None):
+            ep = importlib.metadata.entry_points()
+            if name:
+                return tuple(n for n in ep.get(group, ()) if n.name == name)
+            return ep.get(group, ())
+
+else:
+    import pkg_resources
+
+    def entrypoints(group, name=None):
+        return tuple(pkg_resources.iter_entry_points(group, name))
 
 
 class Registry(object):
@@ -74,10 +98,7 @@ class Registry(object):
         """`list`: the list of supported protocols."""
         _protocols = list(self._protocols)
         if self.load_extern:
-            _protocols.extend(
-                entry_point.name
-                for entry_point in pkg_resources.iter_entry_points("fs.opener")
-            )
+            _protocols.extend(n.name for n in entrypoints("fs.opener"))
             _protocols = list(collections.OrderedDict.fromkeys(_protocols))
         return _protocols
 
@@ -101,10 +122,9 @@ class Registry(object):
         """
         protocol = protocol or self.default_opener
 
-        if self.load_extern:
-            entry_point = next(
-                pkg_resources.iter_entry_points("fs.opener", protocol), None
-            )
+        ep = entrypoints("fs.opener", protocol)
+        if self.load_extern and ep:
+            entry_point = ep[0]
         else:
             entry_point = None
 
--- a/tests/test_opener.py
+++ b/tests/test_opener.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 import sys
 
 import os
-import pkg_resources
 import shutil
 import tempfile
 import unittest
@@ -20,6 +19,11 @@ try:
     from unittest import mock
 except ImportError:
     import mock
+
+if sys.version_info >= (3, 8):
+    import importlib.metadata
+else:
+    import pkg_resources
 
 
 class TestParse(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


eli-schwartz:
@althonos I have resurrected #553 in spirit with what I hope is a universally satisfying and backwards-compatible change. What do you think? Please also see some of my other (recent) PRs which improve the testsuite. It would be great to have some of these merged so that we could have a reliable python 3.12 experience.

MorningLightMountain713:
Can we please get these fixes merged...

eli-schwartz:
I haven't given up on this PR by the way -- I am not aware of any issues with it though, so I'm just waiting for someone with commit access to review...

feanil:
@althonos would it be possible to get a review of this and https://github.com/PyFilesystem/pyfilesystem2/pull/590 ?

eli-schwartz:
I remain committed to seeing this and other PRs go through. As a Linux distro maintainer I consider it my duty as a downstream repackager to do what I can to ensure upstream is healthy as well.

I do lack any ability to merge, release, or otherwise do anything to fix issues other than propose PRs, but I am happy to do that for whatever it's worth...

eli-schwartz:
@lurch @althonos @willmcgugan,



Today's setuptools release has deleted pkg_resources entirely, and pyfilesystem2 is (as a result) completely and utterly broken as it cannot be imported.

What can I do to help you merge my PRs? :)

borisuu:
I was about to get working on a PR for the replacement of pkg_resources. I was pleasantly surprised someone else has already done the work, so let's get this merged and keep this project up-to-date.

shayneoneill:
Any updates on this. Is there a verson of setuptools I can downgrade to in the interim as its blocking me pretty hard at work :(

shayneoneill:
Ok. For anyone looking for this, there is a working version here: https://github.com/milahu/pyfilesystem2/tree/pkg-resources-part-1-and-2

You'll need to mangle your requirements.txt. / Pipfile / etc to point to that specifically. 

Hopefully the maintainer of the official version will merge these fixes in.

edgarrmondragon:
> You'll need to mangle your requirements.txt. / Pipfile / etc to point to that specifically.

You could also just install it from this PR:

```
uv add git+https://github.com/PyFilesystem/pyfilesystem2@refs/pull/589/head
```

yuukigogo000-ai:
Thanks for these — I hit the `pkg_resources` breakage and worked through #589 and #590 together. They do fix the import. One thing I ran into that seems worth catching before this lands:

**`Registry.protocols` stops listing third-party openers on Python 3.10+.**

The 3.10+ branch passes `name` straight through:

```python
def entrypoints(group, name=None):
    ep = importlib.metadata.entry_points(group=group, name=name)
    return tuple(n for n in ep)
```

`Registry.protocols` calls `entrypoints("fs.opener")` without a name, so `name=None` reaches the stdlib. `EntryPoint.matches` compares every keyword with `operator.eq`, so `name=None` matches nothing:

```pycon
>>> import importlib.metadata as im
>>> len(list(im.entry_points(group="console_scripts")))
18
>>> len(list(im.entry_points(group="console_scripts", name=None)))
0
```

The 3.8/3.9 branch is unaffected — `if name:` guards it there. Only the 3.10+ branch.

### Reproduction

master (77a8562) + #590 + #589, built into a wheel and installed into a clean venv, with one third-party opener distribution on the path:

```
master, pkg_resources path (setuptools<81):
  "probe589" in registry.protocols  ->  True    (15 protocols)

+ #590 + #589:
  "probe589" in registry.protocols  ->  False   (14 protocols)
  registry.get_opener("probe589")   ->  ProbeOpener     # still works
  fs.open_fs("probe589://")         ->  works           # still works
```

That is what makes it easy to miss: opening a third-party URL is completely unaffected, so nothing looks broken in normal use. Only enumeration is wrong — including `repr(registry)`, which formats `self.protocols`. Code that does `if proto in registry.protocols` as an input guard will reject a protocol that would have opened fine.

`fs-s3fs` reproduces it the same way if you prefer a real plugin.

### Why the existing test does not catch it

`test_registry_protocols` was written for exactly this behaviour, but it patches `importlib.metadata.entry_points` with a `MagicMock` that returns the extensions regardless of the arguments it is called with, so the `name=None` filtering never happens inside the test.

### Suggested fix

Select on the group only and filter by name in Python:

```python
def entrypoints(group, name=None):
    eps = importlib.metadata.entry_points(group=group)
    return tuple(ep for ep in eps if name is None or ep.name == name)
```

That leaves `get_opener` behaviour identical and restores `protocols`.

### What I actually tested

CPython 3.12.10 on Windows, against #589 at `136e7e2` and #590 at `c009996`, cherry-picked onto master `77a8562` (both applied cleanly), built as a wheel and installed into a fresh venv, then exercised from outside the source tree.

I did **not** test 3.10 or 3.11 — I only have 3.12 here. It is the same code path, and `importlib_metadata` 4.8.3 / 6.0.0 / 7.0.0 / 8.0.0 all return nothing for `name=None`, so I expect them to behave the same, but I have not run them.

One unrelated note for whoever merges this: #589 removes `import pkg_resources` from `tests/test_opener.py` but `test_repr` still calls it, so `tests/test_opener.py` needs #588 applied alongside #589 or it fails with `NameError`.

brianmcgillion:
Thanks for this — together with #590 it's what makes `fs` importable again under setuptools ≥ 82. Two fixups were needed on top; the first is @yuukigogo000-ai's finding above, which I can confirm independently.

**1. `name=None` reaches `importlib.metadata` on 3.10+.** Reproduced against a real plugin (`fs-s3fs`), on Python 3.14:

```
with #589 as-is:      's3' in registry.protocols  ->  False
                      registry.get_opener('s3')   ->  S3FSOpener   # unaffected
```

Their suggested fix restores it:

```diff
         def entrypoints(group, name=None):
-            ep = importlib.metadata.entry_points(group=group, name=name)
-            return tuple(n for n in ep)
+            ep = importlib.metadata.entry_points(group=group)
+            return tuple(n for n in ep if name is None or n.name == name)
```

This was quite quiet: the full suite passes with the bug present, because `test_registry_protocols` mocks `entry_points` with a `MagicMock` that ignores its arguments.

**2. `TestOpeners.test_repr` (line 262) still calls `pkg_resources`.** The new import guard only binds it on Python < 3.8, so this is `NameError: name 'pkg_resources' is not defined` on every supported Python 3:

```diff
     def test_repr(self):
         # Check __repr__ works
-        for entry_point in pkg_resources.iter_entry_points("fs.opener"):
+        for entry_point in importlib.metadata.entry_points(group="fs.opener"):
```

(`registry.entrypoints("fs.opener")` would dogfood the helper and is already imported in that file, but I've only tested the form above.)

With both applied the suite passes on 3.13 and 3.14.

PR #591, opened 2025-01-20
Update test compat with pyftpdlib>=1.5.10

mweinelt:
## Type of changes

- Bug fix

## Checklist

## Description

The pyftpdlib library in version 1.5.10 renamed the `ThreadedTestFTPd` to `FtpdThreadWrapper`. I've updated that reference and ran the tests. I also updated the version requirement to my best knowledge.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -2,7 +2,7 @@
 
 # pyftpdlib is needed to spawn a FTP server for the
 # FTPFS test suite
-pyftpdlib ~=1.5
+pyftpdlib ~=1.5.10
 
 # these are optional dependencies for pyftpdlib that
 # are not explicitly listed, we need to install these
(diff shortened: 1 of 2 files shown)


musicinmybrain:
This seems to be on the right track, but doesn’t seem quite right at the moment. It looks like `pyftpdlib` 1.5.10 still has `ThreadedTestFTPd`, and `FtpdThreadWrapper` was introduced in `pyftpdlib` 2.0.

https://github.com/giampaolo/pyftpdlib/blob/3cea779de3ccb5ab5744220de3bddb3d4ff10f35/pyftpdlib/test/__init__.py#L445

Also, there is another use of `ThreadedTestFTPd` in `TestAnonFTPFS` that needs to be adjusted:

https://github.com/PyFilesystem/pyfilesystem2/blob/77a8562785fc37cb2e30bdcd39c133097ba62dce/tests/test_ftpfs.py#L331-L347

I used this PR as inspiration for https://github.com/PyFilesystem/pyfilesystem2/pull/595.

mweinelt:
Back then I applied a substitution downstream, so in practice I arrived at the same result as #595, and missed that it touched two tests. Thanks!

PR #600, opened 2026-06-17
fix: carry filesize suffix when mantissa rounds up to base

patchwright:
## Problem

All three public functions (`traditional`, `binary`, `decimal`) can produce impossible output like `"1,024.0 KB"` instead of `"1.0 MB"` for values just below a unit boundary.

Root cause: `_to_str` picks the suffix from the *unrounded* value, then formats the mantissa — which can round up to `base`, producing an impossible result.

```python
filesize.decimal(999_999)        # '1,000.0 kB'  ← should be '1.0 MB'
filesize.traditional(1024**2-1)  # '1,024.0 KB'  ← should be '1.0 MB'
filesize.binary(1024**2-1)       # '1,024.0 KiB' ← should be '1.0 MiB'
```

## Fix

Convert `suffixes` to a list before the loop. After computing the mantissa, round and compare against `base`. If the rounded value hits `base`, step up one suffix and recompute:

```python
mantissa = base * size / unit
if round(mantissa, 1) >= base and i - 1 < len(suffixes):
    suffix = suffixes[i - 1]
    mantissa = size / unit
```

Math: when the current unit is `base**i`, dividing `size` by that same `unit` gives the value in the next suffix (whose denominator is `base**(i+1)`, which divides into `base * size / base**(i+1) = size / base**i = size / unit`).

Also removes the existing TODO comment that flagged this exact issue.

## Tests

Added `test_rollover_decimal`, `test_rollover_traditional`, and `test_rollover_binary` covering the unit boundaries for all three functions.

### DIFF
--- a/fs/filesize.py
+++ b/fs/filesize.py
@@ -33,12 +33,20 @@ def _to_str(size, suffixes, base):
     elif size < base:
         return "{:,} bytes".format(size)
 
-    # TODO (dargueta): Don't rely on unit or suffix being defined in the loop.
+    suffixes = list(suffixes)
     for i, suffix in enumerate(suffixes, 2):  # noqa: B007
         unit = base**i
         if size < unit:
             break
-    return "{:,.1f} {}".format((base * size / unit), suffix)
+
+    mantissa = base * size / unit
+    # Rounding can push the mantissa to `base` (e.g. 999,999 B is 999.999 kB,
+    # which formats as "1,000.0 kB"). Carry into the next suffix instead.
+    if round(mantissa, 1) >= base and i - 1 < len(suffixes):
+        suffix = suffixes[i - 1]
+        mantissa = size / unit
+
+    return "{:,.1f} {}".format(mantissa, suffix)
 
 
 def traditional(size):
--- a/tests/test_filesize.py
+++ b/tests/test_filesize.py
@@ -60,6 +60,15 @@ class TestFilesize(unittest.TestCase):
         self.assertEqual(filesize.binary(1024**2 - 1), "1.0 MiB")
         self.assertEqual(filesize.binary(1024**3 - 1), "1.0 GiB")
 
+    def test_rounding_boundaries(self):
+        # Within-unit rounding boundaries (no suffix change), per review on #600
+        self.assertEqual(filesize.traditional(1996), "1.9 KB")
+        self.assertEqual(filesize.traditional(1997), "2.0 KB")
+        self.assertEqual(filesize.decimal(1950), "1.9 kB")
+        self.assertEqual(filesize.decimal(1951), "2.0 kB")
+        self.assertEqual(filesize.binary(1996), "1.9 KiB")
+        self.assertEqual(filesize.binary(1997), "2.0 KiB")
+
     def test_errors(self):
 
         with self.assertRaises(TypeError):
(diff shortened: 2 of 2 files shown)


lurch:
Ahhh, the fun of floating-point rounding :laughing: 

I wonder if it might also be worth adding some unit-tests for the boundary-values? E.g.
```python
>>> traditional(1996)
'1.9 KB'
>>> traditional(1997)
'2.0 KB'
>>> decimal(1950)
'1.9 kB'
>>> decimal(1951)
'2.0 kB'
>>> binary(1996)
'1.9 KiB'
>>> binary(1997)
'2.0 KiB'
```

patchwright:
Good call, those within-unit boundary cases are worth pinning down separately from the rollover ones. Added a `test_rounding_boundaries` covering your exact examples for all three functions:

```python
traditional(1996) == "1.9 KB"
traditional(1997) == "2.0 KB"
decimal(1950)     == "1.9 kB"
decimal(1951)     == "2.0 kB"
binary(1996)      == "1.9 KiB"
binary(1997)      == "2.0 KiB"
```

All pass on the branch. Thanks for the review.

lurch:
Is it also worth checking the boundaries at e.g.
```python
>>> decimal(999_949)
'999.9 kB'
>>> decimal(999_950)
'1.0 MB'
```
or do you think those are sufficiently covered by the existing tests?

patchwright:
Good catch. Those weren't covered: `test_rollover_decimal` only pinned the far end at `999_999`, and `test_rounding_boundaries` is within-unit, so the first-flip at `999_950` had no test. Both behave as expected: `decimal(999_949)` stays `'999.9 kB'`, `decimal(999_950)` rolls to `'1.0 MB'`.Added that pair. `traditional` and `binary` share the same carry logic and had the same gap (their rollover tests only checked `1024**2 - 1`), so I pinned their first-flip too: `1_048_524` stays `'1,023.9 KB'`/`'1,023.9 KiB'`, `1_048_525` rolls to `'1.0 MB'`/`'1.0 MiB'`. Pushed in the latest commit, full suite green.

patchwright:
FYI on the red checks: the failing jobs (test (2.7), (3.5) through (3.10), (pypy-2.7/3.6/3.7)) all fail identically on master @77a8562, so they are pre-existing failures on the EOL-Python jobs rather than anything from this diff. As a small upside, `lint (typecheck)` is failing on master but passes on this branch. The rollover tests added here pass on supported Pythons (3.10+).

patchwright:
Quick correction to my last CI note: I called those red `test (...)` jobs EOL-Python failures matching master. Looking at the check annotations, that is not quite right. GitHub is auto-failing every test job because `.github/workflows/test.yml` (line 45) uses `actions/upload-artifact@v2`, which was deprecated and auto-fails per the 2024-02-13 policy. It is not from this diff: the lint jobs pass (including typecheck); only the test jobs hit the deprecated-action fail. Bumping `upload-artifact` to v4 would clear the matrix. I will leave that to you since it is CI infra, not the filesize change.

patchwright:
Checking back in — it's been about a month. The content side (filesize suffix + the boundary tests you asked for) has been stable since 07-07; the only red is the pre-existing `upload-artifact@v2` deprecation on the test matrix (unrelated to this diff, lint jobs all pass). Let me know if there's anything else you'd like changed.

